PR #9, opened 2022-06-27
Add support for all FERC forms

zschira:
This pull request adds support for selecting any Ferc form taxonomy for constructing the output database. It also includes some minor refactoring of how XBRL instances are parsed. Now it will return a class which wraps the parsed `Facts` and `Contexts`, rather than returning them directly.

### DIFF
--- a/src/ferc_xbrl_extractor/cli.py
+++ b/src/ferc_xbrl_extractor/cli.py
@@ -11,7 +11,13 @@ from ferc_xbrl_extractor import helpers, xbrl
 from ferc_xbrl_extractor.helpers import get_logger
 from ferc_xbrl_extractor.instance import InstanceBuilder
 
-DEFAULT_TAXONOMY = "https://eCollection.ferc.gov/taxonomy/form1/2022-01-01/form/form1/form-1_2022-01-01.xsd"
+TAXONOMY_MAP = {
+    1: "https://eCollection.ferc.gov/taxonomy/form1/2022-01-01/form/form1/form-1_2022-01-01.xsd",
+    2: "https://eCollection.ferc.gov/taxonomy/form2/2022-01-01/form/form2/form-2_2022-01-01.xsd",
+    6: "https://eCollection.ferc.gov/taxonomy/form6/2022-01-01/form/form6/form-6_2022-01-01.xsd",
+    60: "https://eCollection.ferc.gov/taxonomy/form60/2022-01-01/form/form60/form-60_2022-01-01.xsd",
+    714: "https://eCollection.ferc.gov/taxonomy/form714/2022-01-01/form/form714/form-714_2022-01-01.xsd",
+}
 
 
 def parse_main():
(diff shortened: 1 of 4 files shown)


src/ferc_xbrl_extractor/cli.py
```diff
@@ -9,9 +9,15 @@
...
-from ferc_xbrl_extractor.instance import Instance
+from ferc_xbrl_extractor.instance import InstanceBuilder
 
-DEFAULT_TAXONOMY = "https://eCollection.ferc.gov/taxonomy/form1/2022-01-01/form/form1/form-1_2022-01-01.xsd"
+TAXONOMY_MAP = {
+    1: "https://eCollection.ferc.gov/taxonomy/form1/2022-01-01/form/form1/form-1_2022-01-01.xsd",
+    2: "https://eCollection.ferc.gov/taxonomy/form2/2022-01-01/form/form2/form-2_2022-01-01.xsd",
+    6: "https://eCollection.ferc.gov/taxonomy/form6/2022-01-01/form/form6/form-6_2022-01-01.xsd",
+    60: "https://eCollection.ferc.gov/taxonomy/form60/2022-01-01/form/form60/form-60_2022-01-01.xsd",
+    714: "https://eCollection.ferc.gov/taxonomy/form714/2022-01-01/form/form714/form-714_2022-01-01.xsd",
```
bendnorman, line 19: Does FERC plan to release taxonomies that aren't just integers?

bendnorman: Also, will these taxonomies change? Are we archiving them?

src/ferc_xbrl_extractor/instance.py
```diff
@@ -253,3 +200,125 @@ def from_xml(cls, elem: Element) -> "Fact":
...
+        """
+        if instant:
+            facts = self.instant_facts.get(tuple(axes), {})
+        else:
+            facts = self.duration_facts.get(tuple(axes), {})
+
+        return facts
+
+
+class InstanceBuilder:
```
bendnorman, line 267: What is the benefit of adding an `InstanceBuilder` over keeping the parsing logic within `Instance`?

src/ferc_xbrl_extractor/cli.py
zschira: I'm not really sure if they will ever have any non-integer forms to worry about, but as of now they all are, so I just did it this way for simplicity.

The taxonomies do change, but it would be a new version that would be have a different URL. Every XBRL filing contains a reference to the taxonomy that was used to construct it, so the URL's here _should_ point to the 2022 version of the taxonomy forever. That being said, we probably should archive the taxonomies just to be safe.

src/ferc_xbrl_extractor/instance.py
bendnorman: Ah got it. Thanks!

bendnorman (approved):
Looks ready to go! I think we should create a separate issue for archiving XBRL taxonomies. They could probably just be saved in this git repo instead of zenodo.

PR #13, opened 2022-08-05
Determine mapping between FERC 1 DBF `respondent_id` and XBRL `entity_id`

zschira:
This PR includes a notebook used for developing a mapping between FERC `respondent_id`'s and `entity_id`'s (also called Company Identifiers or CID's). This mapping will ultimately be added to PUDL, but the purpose of this PR is to verify the methodology for creating this map in the notebook in question.


zschira:
Yeah this should be closed

PR #15, opened 2022-08-19
Enable publication of raw extracted XBRL data

zschira:
This PR adds various improvements to the XBRL extractor to enable the archiving/publication of raw SQLite DB's extracted from XBRL data.

Specifically, it:

- Adds the ability to specify a subset of tables for extraction
- Makes column names snake_case to generate valid frictionless datapackages
- Fixes broken documentation generation


README.rst
```diff
@@ -19,11 +19,15 @@ Installation
```
bendnorman, line 19: Any reason you are using .rst instead of .md?

src/ferc_xbrl_extractor/instance.py
```diff
@@ -230,13 +230,15 @@ def __init__(
             filing_name: Name of parsed filing.
         """
         # This is a nested dictionary of dictionaries to locate facts by context
-        self.instant_facts = {}
-        self.duration_facts = {}
+        self.instant_facts: dict[tuple, dict[Context, list[Fact]]] = {}
+        self.duration_facts: dict[tuple, dict[Context, list[Fact]]] = {}
```
bendnorman, line 234: What a type hint! What are the keys of these dicts?

src/ferc_xbrl_extractor/taxonomy.py
```diff
@@ -149,29 +161,35 @@ def from_list(cls, linkrole_list: List, concept_dict: ConceptDict) -> "LinkRole"
...
+        # Create dictionary mapping concept names to concepts
         concept_dict = {
             str(name): concept for name, concept in taxonomy.qnameConcepts.items()
         }
 
-        # References provide sometimes useful metadata that can be saved in JSON
-        if metadata_fname:
-            save_references(metadata_fname, concept_dict)
-
         roles = [
```
bendnorman, line 193: Why was `save_references()` removed?

bendnorman (approved):

README.rst
zaneselvans: I think this is because the docs build from `cheshire` assumes everything is RST.

src/ferc_xbrl_extractor/instance.py
zaneselvans: Whew yeah, that looks like it might be a candidate for a typed dict or Pydantic class or something.

zschira: I've added a type alias with some explanation of how it's used. I'm not sure that a pydantic class or typed dict would be useful because the keys are pretty dynamic. Let me know if this clarifies things at all

src/ferc_xbrl_extractor/taxonomy.py
zschira: At this point `save_references` was really an artifact of some earlier work that's not actually in use now. XBRL references contain additional metadata that I thought would be useful for linking XBRL data to DBF data, but it didn't end up working how I thought it might, and I figured at this point it would cause more confusion to leave this in.

README.rst
zschira: Yeah I'm just following the patterns from `cheshire` here

src/ferc_xbrl_extractor/taxonomy.py
```diff
@@ -1,19 +1,20 @@
 """XBRL prototype structures."""
-from typing import Dict, ForwardRef, List, Literal
+from __future__ import annotations
```
zaneselvans, line 2: Have you tried running this on Python 3.8 / 3.9? I found that the combination of Pydantic's runtime type checking magic and the `__future__` annotations broke things in some cases.

tox.ini
```diff
@@ -60,6 +61,15 @@ extras =
...
     bandit -r src/ferc_xbrl_extractor/
 
+[testenv:mypy]
+description = Run mypy type checking on the codebase.
+skip_install = false
+extras =
+    tests
+    types
+commands =
+    mypy --strict src tests
```
zaneselvans, line 71: Wowsers, did you really get it all working with `--strict`?

src/ferc_xbrl_extractor/datapackage.py
```diff
@@ -405,3 +382,46 @@ def construct_dataframe(self, instance: Instance) -> pd.DataFrame:
...
+
+class Datapackage(BaseModel):
+    """A generic Data Package, as per Frictionless Data specs.
+
+    See https://specs.frictionlessdata.io/data-package.
+    """
+
+    profile: str = "tabular-data-package"
+    name: str = "ferc1-extracted-xbrl"
+    title: str = "Ferc1 data extracted from XBRL filings"
```
zaneselvans, line 395: Is this actually specific to the FERC 1? Or will the name and title need to be set dynamically based on which form is being extracted?

src/ferc_xbrl_extractor/cli.py
```diff
@@ -32,9 +32,9 @@ def parse_main():
     )
     parser.add_argument(
         "-s",
-        "--save-metadata",
+        "--save-datapackage",
         default=None,
-        help="Save metadata defined in XBRL references",
+        help="Generate frictionless datapackage descriptor, and write to JSON file at specified path.",
```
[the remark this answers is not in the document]
zschira: I currently have an integration test that does exactly that. Do you think they should be validated as a part of the of the actual tool, and not just during testing?

src/ferc_xbrl_extractor/cli.py
```diff
@@ -102,8 +102,8 @@ def get_instances(instance_path: Path):
     # Directory of instances
     else:
         # Must be either a directory or file
-        assert instance_path.is_dir()
-        instances = instance_path.iterdir()
+        assert instance_path.is_dir()  # nosec: B101
+        instances = list(instance_path.iterdir())
```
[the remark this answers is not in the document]
zschira: Fixed

src/ferc_xbrl_extractor/taxonomy.py
zschira: I just tested on python 3.8 and 3.9 and I am getting errors as well. Have you found a solution for this?

src/ferc_xbrl_extractor/datapackage.py
zschira: Good catch, it was just using FERC 1 when it should be dynamically set. Fixed now

src/ferc_xbrl_extractor/taxonomy.py
zaneselvans: I think the two options are to either not upgrade the type hints, or to only use the most recent version of Python. We went with only using Python 3.10 on the main pudl repo.

src/ferc_xbrl_extractor/cli.py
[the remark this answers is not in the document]
zaneselvans: I don't think it takes very long and I suspect it'll catching things sometimes that don't come up in the tests.

README.rst
zaneselvans: I think this package is generically useful enough that it would be good to write some basic docs explaining what it's doing and publish on RTD, so I think this is the right move.

zaneselvans:
Let me know when this does get merged so I can update my local XBRL DB with the new column names and integrate them into the transform module.

src/ferc_xbrl_extractor/cli.py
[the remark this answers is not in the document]
zschira: Yeah I feel like we'll be more likely to catch any errors resulting from year-to-year changes in the taxonomy, by directly validating the descriptor, so I've added a check here.

src/ferc_xbrl_extractor/taxonomy.py
zschira: Ok, I've done the same as PUDL here. I can definitely change that if we have a compelling reason to support python 3.8/9, but I don't personally see one right now

src/ferc_xbrl_extractor/cli.py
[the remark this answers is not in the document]
zschira: This has actually already proven to be a good call. I just found an error due to some special characters in FERC 2 column names that I wasn't catching earlier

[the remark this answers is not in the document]
zaneselvans: Yass! Yay tests.

src/ferc_xbrl_extractor/taxonomy.py
zaneselvans: Sounds good. Most of this is much more "application" like, and I think in that case just the most recent version will make our lives easier.

src/ferc_xbrl_extractor/datapackage.py
zaneselvans: Are these just default values now that are dynamically overridden?  Should there be default values at all if they're dynamically generated?

zaneselvans (commented):
Sorry, a few lingering questions!

src/ferc_xbrl_extractor/datapackage.py
```diff
@@ -182,9 +201,8 @@ def _lowercase_words(name: str) -> str:
     return name
 
 
-def _clean_table_names(name: str) -> Optional[str]:
-    """
-    Function to clean table names.
+def _clean_table_names(name: str) -> str | None:
```
zaneselvans, line 204: Maybe this is all clean enough now that it doesn't matter, but if you find yourself chasing down all manner of different special characters, it turns out there are some string normalization functions built into the standard Python libraries, and Python validators that mutate the value during validation seem like they might be a good fit for taking an existing table name and outputting strict snake case.  I think there's a `SnakeCase` model in `pudl.metadata.classes` but don't know if it does what you need.

src/ferc_xbrl_extractor/datapackage.py
```diff
@@ -198,37 +216,41 @@ def _clean_table_names(name: str) -> Optional[str]:
...
-    )
+
+    # Convert to snakecase
     table_name = stringcase.snakecase(table_name)
 
+    # Remove all special characters
+    table_name = re.sub(r"\W", "", table_name)
+
     # The conversion to snakecase leaves some names with multiple underscores in a row
     table_name = table_name.replace("___", "_").replace("__", "_")
```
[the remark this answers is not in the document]
zschira: Yeah I realized the same thing and changed this

src/ferc_xbrl_extractor/datapackage.py
```diff
@@ -1,20 +1,21 @@
 """Define structures for creating a datapackage descriptor."""
+from __future__ import annotations
+
```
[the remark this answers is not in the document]
zschira: Removed

src/ferc_xbrl_extractor/datapackage.py
zschira: Yeah the default isn't doing anything now. I've removed it

src/ferc_xbrl_extractor/datapackage.py
zschira: Hmm I just read up on the pydantic `constr` [here](https://pydantic-docs.helpmanual.io/usage/types/#arguments-to-constr) used for `SnakeCase`, and it seems like it can't quite handle everything I need. From my understanding it can do things like convert to lowercase and strip whitespace, but the regex used for `SnakeCase` is only used for validation, and doesn't actually do any normalization. I'm not 100% sure I'm reading the docs correctly though

zaneselvans: Ah okay, that sounds right that our SnakeCase is just validating, and not mutating.

The standard library unicode normalization function I recently came across was [unicodedata.normalize](https://docs.python.org/3/library/unicodedata.html#module-unicodedata).

Sounds like there's nothing off the shelf that will do what's necessary here, so looks good to me.

zaneselvans (approved):

zaneselvans:
Mrrrrrrge me!

PR #22, opened 2022-09-23
Dev

zschira:
This pull request adds the ability to parse XBRL taxonomies from a zipfile. Arelle already has this capability, so it's a fairly small change to tell Arelle the location of the zipfile, and the path to the taxonomy entry point within it.

### DIFF
--- a/src/ferc_xbrl_extractor/arelle_interface.py
+++ b/src/ferc_xbrl_extractor/arelle_interface.py
@@ -38,7 +38,6 @@ def load_taxonomy_from_archive(filepath: str, archive_path: str):
         archive_path: Relative path to taxonomy entry point within archive.
     """
     # Create arelle FileSource object
-    print(f"archive_path={archive_path}")
     f = FileSource.openFileSource(archive_path, sourceZipStream=filepath)
 
     return load_taxonomy(f)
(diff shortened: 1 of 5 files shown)


src/ferc_xbrl_extractor/arelle_interface.py
```diff
@@ -28,3 +28,17 @@ def load_taxonomy(path: str):
...
+
+def load_taxonomy_from_archive(filepath: str, archive_path: str):
+    """Load an XBRL taxonomy from a zipfile archive.
+
+    Args:
+        filepath: Path to zipfile on disc.
+        archive_path: Relative path to taxonomy entry point within archive.
+    """
+    # Create arelle FileSource object
+    print(f"archive_path={archive_path}")
```
bendnorman, line 41: Can this be replaced with a logger method?

bendnorman (changes requested):
Looks good! So before these changes, the `xbrl_extract ` command extracted XBRL filing directly from the FERC API. Now, you can point `--archive-path` to the archived XBRL filings in zipfile produced by your scrapers?

src/ferc_xbrl_extractor/arelle_interface.py
zschira: Oh yeah I'd just put this in for debugging, so I just removed it

zschira:
Previously it was getting the taxonomies from FERC directly, but it was parsing taxonomies from disk. This option allows it to parse taxonomies from a local zipfile.

bendnorman (approved):

PR #29, opened 2022-11-02
Add function to output metadata from taxonomies to json file

zschira:
This PR adds the ability to extract additional metadata from taxonomies that can be used for mapping XBRL -> DBF data. For now all of this metadata is simply dumped in a JSON file that allows looking up any desired values. The JSON structure looks something like:
```
{
    ...,
    "electric_plant_in_service": {
        "references": {
            "Form Location": [
                {
                    "Form":  "Form 1",
                    "Schedule": "204 - Schedule - Electric Plant In Service",
                    ...
                },
                ...
            ]
        },
        "calculations": [
            {
                "name": "electric_plant_in_service_and_completed_construction_not_classified_electric",
                "weight": 1.0,
            },
            ...
        ],
        "balance": "debit",
    }
}
```


zaneselvans (approved):
No comments on the output format -- if it's got all the the taxonomy info and it's machine readable, which seems to be the case, then we're good for now!

I do think we should go through all the command line options and try to standardize them a bit. Similar naming seems to mean different things in different cases.  And I think it would be really helpful to have some examples of command line usage -- I couldn't tell from the help message how to actually use it, find appropriate inputs, etc.

While it makes sense to have the flexibility to read the inputs from anywhere and write the outputs to anywhere in this generic tool, I we should come up with a convention that's used in the context of PUDL, like we do for where the data is currently read from. I guess for now defualting to writing the JSON out alongside the datapackage.json and ferc1-xbrl.sqlite in the same directory makes sense. With a more descriptive name than just "metadata.json".

Also, is there a good reason not to output the JSON on multiple lines with indentation for human readability? I had to reformat it to poke around in there. I guess it's a bigger file because whitespace. I guess we can reformat and browse in VS Code and just try to be careful not to save.

zschira:
Yeah the CLI could definitely use some improvement/standardization. It's grown a lot without having a ton of thought put into making it clear and easy to use.

As for the formatting, I've changed it so it should be on multiple lines with indentation, so hopefully that helps.

Issue #33, opened 2022-11-17
Group extracted XBRL taxonomy metadata by table


zaneselvans:
Right now all of the XBRL taxonomy metadata we extract is organized into an array of JSON objects, with each one pertaining to a single XBRL fact. E.g. in the context of the XBRL derived `electric_plant_in_service` table there's a top level object for each of the following FERC accounts or aggregations of FERC accounts (as well as additional facts for each one with suffixes `_additions`, `_retirements`, `_adjustments`, and `_transfers`):

* `land_and_land_rights_steam_production`
* `structures_and_improvements_steam_production`
* `boiler_plant_equipment_steam_production`
* `engines_and_engine_driven_generators_steam_production`
* `turbogenerator_units_steam_production`
* `accessory_electric_equipment_steam_production`
* `miscellaneous_power_plant_equipment_steam_production`
* `steam_production_plant`

But there's no programmatically usable indication that each one of these array elements belongs to the `electric_plant_in_service` table.

This complicates our use of the metdata in the ETL pipeline, since it means we can't select the appropriate subset of the metadata fields for use with a given table unless we already have a list of all of the columns that appear in that table. Right now we're reading the column names at the beginning of the XBRL table transform, and then using them to grab the right metadata, but it would be much simpler if we could grab the right metadata at the beginning, knowing only the name of the table (i.e. `electric_plant_in_service`).

Given how generic some of the field names are (like `organization`...) it's also kind of surprising that they would be globally unique or always have the same meaning among the thousands of FERC 1 fields.

Clearly there's information in the XBRL taxonomies that indicate what fields are part of which tables. How can we integrate that into the taxonomy export so that we can select one table's worth of metadata at a time without needing to extract a list of column names from the actual database first?

Would it make sense to make it a dictionary of dictionaries, each of which contains an array of "facts"? E.g.

```json
{
  "electric_plant_in_service_204": {
    [
      "land_and_land_rights_steam_production": {...},
      "structures_and_improvements_steam_production": {...},
      "boiler_plant_equipment_steam_production": {...},
      "engines_and_engine_driven_generators_steam_production": {...},
      "turbogenerator_units_steam_production": {...},
      "accessory_electric_equipment_steam_production": {...},
      "miscellaneous_power_plant_equipment_steam_production": {...},
      "steam_production_plant": {...},
      ...
    ]
  },
  "electric_energy_account_401a": {
    [
      ...
    ]
  }
}
```

[...]

PR #38, opened 2022-12-01
Group extracted XBRL taxonomy metadata by table

zschira:
This PR addresses #33 to group taxonomy metadata by table. It also adds pydantic models wrapping the metadata structures to formalize the extraction, and improve readability.


zaneselvans (changes requested):
Hey Zach, I've spent an hour with this and am still having trouble understanding what's going on and how it's different from what was happening before.

I think we need to add an XBRL glossary to the docs -- there are a ton of really generic words coming from the XBRL nomenclature that don't mean anything without XBRL specific context that I think only you have. We need to connect those words to our use cases and the structures that we're working with to make this more legible.

* What thing corresponds to a table in the database that's being generated?
* What thing corresponds to a column in the database that's being generated?
* Where is it identifying groups of columns that are part of the same table?

Some terms that should go in the glossary:

* Role
* Concept
* Concept Tree
* Link Role
* Entity
* Fact
* Fact Table
* Instance
* Context
* Period

We also need concrete examples in the README -- commands with real arguments that someone could actually run and have the extraction script do something. Like this, but filled in, and probably pointing at a local archive including a taxonomy, rather than a URL:

```
$ xbrl_extract {path_to_filings} {path_to_database} --taxonomy {url_of_taxonomy}
```

We can check one of the smaller XBRL archive files into the repo for testing and usage examples. `ferc6-xbrl-2020.zip` is only 800k.

zschira:
@zaneselvans yeah I think creating a glossary of terms is a very good idea. The nomenclature is really confusing. I also think I should create some diagrams to illustrate the extraction process, because I continually struggle to effectively describe it in words.

Generally I've been wanting to dedicate some time to improving the extractor repo with things like tests, CI, overhauling the CLI and docs to make it easier to use etc.

README.rst
```diff
@@ -42,14 +42,32 @@ path to the SQLite database.
...
 the structure of the output database. To specify a different taxonomy use the
-``--taxonomy`` option.
+``--taxonomy`` option. This can be either a URL pointing to a taxonomy entry point,
+or a path to a local zipfile.
 
 .. code-block:: console
 
     $ xbrl_extract {path_to_filings} {path_to_database} --taxonomy {url_of_taxonomy}
 
+If using a zipfile, you must also specify the path to the taxonomy entry point within
```
zaneselvans, line 61: "If you're using a zipfile..." for what purpose? To supply the filings? Or to supply the taxonomy? Or both?

Can you use a zipfile for the filings, and a URL for the taxonomy? Or if you're using a zipfile for the filings, do you **have** to use a path to the taxonomy within the zipfile?

If reading filings from within the zipfile means you *must* use a taxonomy from within the zipfile, then would it make sense to just use `--taxonomy` to specify either the URL (if not using a zipfile) or the path within the archive (if using a zipfile)?

Then once there's a standardized location within the zipfile for the taxonomy, the UI for both the "files / directory + URL" and the "zipfile + archived taxonomy" would be the same -- with no `--taxonomy` option, they'd just use the default location (either URL or archive).

README.rst
```diff
@@ -42,14 +42,32 @@ path to the SQLite database.
 
     $ xbrl_extract {path_to_filings} {path_to_database}
 
+This repo contains a small selection of FERC Form 1 filings from 2021 in the
+``examples`` directory. To test the tool on these filings, use the command
+
+.. code-block:: console
+
+    $ xbrl_extract examples/ferc1-2021-sample.zip {path_to_database}
```
zaneselvans, line 50: Can you go ahead and include a dummy value for `path_to_database` here for clarity, e.g.

```
$ xbrl_extract examples/ferc1-2021-sample.zip ./ferc1-2021-sample.sqlite
```

Is this a database which has to already exist? Or is this a new database that will be created? What happens if a database already exists at that location?

README.rst
```diff
@@ -62,3 +80,15 @@ optimally configured.
...
+First is the ``--save-datapackage`` command to save a
+`frictionless datapackage <https://specs.frictionlessdata.io/data-package/>`__ as a
+json file. There is also the ``--metadata-path``, which writes taxonomy metadata to
+a json file that groups metadata by table name. See the ``arelle_interface`` module
+for more info on the extracted metadata. To create both of these files using the
+example filings and taxonomy, run the following command.
+
+.. code-block:: console
+
+    $ xbrl_extract examples/ferc1-2021-sample.zip {path_to_database} --taxonomy examples/taxonomy.zip --archive-path taxonomy/form1/2021-01-01/form/form1/form-1_2021-01-01.xsd --metadata-path metadata.json --save-datapackage datapackage.json
```
zaneselvans, line 94: Let's break this into a multi-line command for readability, and use the same file naming conventions that we have in the PUDL repo, so that they feel familiar if someone runs across them both:

```
$ xbrl_extract examples/ferc1-2021-sample.zip ferc1-2021-sample.sqlite \
    --taxonomy examples/taxonomy.zip \
    --archive-path taxonomy/form1/2021-01-01/form/form1/form-1_2021-01-01.xsd \
    --metadata-path ferc1_xbrl_taxonomy_metadata.json \
    --save-datapackage ferc1_xbrl_datapackage.json
```

This command surprised me. I did not understand earlier on that you were talking about two (potentially?) different zipfiles for the filings themselves and the taxonomy. Do they have to be separate? Are they allowed to be in the same zipfile? How is it arranged in the PUDL datastore archives?

README.rst
```diff
@@ -42,14 +42,32 @@ path to the SQLite database.
...
     $ xbrl_extract {path_to_filings} {path_to_database}
 
+This repo contains a small selection of FERC Form 1 filings from 2021 in the
+``examples`` directory. To test the tool on these filings, use the command
+
+.. code-block:: console
+
+    $ xbrl_extract examples/ferc1-2021-sample.zip {path_to_database}
+
 By default, the CLI will use the 2022 version of the FERC Form 1 Taxonomy to create
```
zaneselvans, line 52: Is it specifically using the **2022** taxonomy by default? Or the **current year**? If it's specifically 2022, I imagine we would want to update it each year.

docs/extract.rst
```diff
@@ -1,51 +1,51 @@
...
-leaf node. After this process is complete, we can parse individual filings, and
-populate these constructed tables with the data in the filing.
+`Arelle <https://arelle.org/arelle/>`__ is a popular open source library for working
+with XBRL data, and it is used for parsing the taxonomy. Arelle can parse a taxonomy
+from either a URL to the taxonomy entry point, or a local path to a zipfile containing
+the entire taxonomy.
+
+The extractor will first take a parsed taxonomy and construct a Taxonomy object defined
+in ``ferc_xbrl_extractor/taxonomy.py``. This is done because the data structures used by
+Arelle are not well documented, so it's immediately move to costum structures. After
```
zaneselvans, line 14: Typos.

docs/extract.rst
```diff
@@ -1,51 +1,51 @@
...
+creating these structures, the extractor will generate
+`frictionless datapackage <https://specs.frictionlessdata.io/data-package/>`__, which
+contains a schema for the new SQLite DB, as well as useful metadata.
```
zaneselvans, line 17: **a** frictionless datapackage.

Is it a generic datapackage, or is it a tabular datapackage? If it's the more specific tabular type, let's link to the more specific specification.

docs/extract.rst
```diff
@@ -1,51 +1,51 @@
...
+
+Each table in the SQLite DB is derived from a Link Role in the Taxonomy. This is done
```
zaneselvans, line 19: I think linking to the glossary from the terms in this document would be helpful. Sphinx has [built-in support](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-glossary) term references / glossaries.

docs/glossary.rst
```diff
@@ -0,0 +1,66 @@
+===============================================================================
+XBRL Glossary
```
zaneselvans, line 2: Sphinx has a built-in `glossary::` directive that will be helpful here, and make it easy to reference the defined terms. [See the Sphinx docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-glossary)

docs/glossary.rst
```diff
@@ -0,0 +1,66 @@
...
+===============================================================================
+
+Axis
+^^^^
+
+Axes are defined in the Taxonomy, and are used as a part of the
```
zaneselvans, line 8: Need to explain how the Axes relate to primary keys in the database tables that end up getting produced. An example would be helpful.

docs/glossary.rst
```diff
@@ -0,0 +1,66 @@
...
+Context to identify a Fact beyond the default
+entity, and Period. Axes can be either typed or explicit. An
+explicit Axis will have all of its allowable values defined in the taxonomy. A
+typed Axis will instead allow a filer to input their own values for the Axis.
+
+Concept
```
zaneselvans, line 14: What does an XBRL Concept correspond to in the extracted relational database structure? Is it a data type? Is it a variable that's being reported across many filings in a single table? Can the same Concept appear in multiple tables, and if it does will it always have the same meaning?

docs/glossary.rst
```diff
@@ -0,0 +1,66 @@
...
+^^^^^^^
+
+Concepts are defined in the Taxonomy, and describe what sort of Facts can be
+reported in a Filing. They provide a written description, and other metadata that
+the taxonomy creator chooses to include. Not all Concepts will have Facts they are
+directly associated with. This is because Concepts can also be used as a part of a
+Concept tree to define relationships with other Concepts. The root of such a tree
+is a Link Role.
+
+Context
```
zaneselvans, line 24: Is this purely a metadata structure? Or does it correspond to some data structure that would appear in the extracted relational DB?

docs/glossary.rst
```diff
@@ -0,0 +1,66 @@
...
+
+Concept
+^^^^^^^
+
+Concepts are defined in the Taxonomy, and describe what sort of Facts can be
+reported in a Filing. They provide a written description, and other metadata that
+the taxonomy creator chooses to include. Not all Concepts will have Facts they are
+directly associated with. This is because Concepts can also be used as a part of a
+Concept tree to define relationships with other Concepts. The root of such a tree
+is a Link Role.
```
zaneselvans, line 22: Definitely need an example for this nested / tree structure. I thought a Link Role defined a table, in which case I would have thought it would always have data inside it.

docs/glossary.rst
```diff
@@ -0,0 +1,66 @@
...
+
+Context
+^^^^^^^
+
+A context provides important information for identifying/interpreting a
+Fact. Every context contains both an entity identifier, and a
+Period. A context can also contain user defined Axes.
+
+Fact
```
zaneselvans, line 31: What does a Fact correspond to in the extracted relational database structure? Is it a whole record? A whole column? A single value at the intersection of a row and a column?

docs/glossary.rst
```diff
@@ -0,0 +1,66 @@
...
+^^^^
+
+An XBRL fact is considered to be an atomic unit of data. It contains a single value,
+with a name that corresponds to a Concept, and a reference to a
+Context.
+
+Filing
```
zaneselvans, line 38: What does this mean in the context of the FERC data? Each filing is all of the data reported by a single utility (respondent / entity) in a single year (or other reporting period), so it contains information that will end up being turned into one or more rows in many (or even all) of the different tables.

Can we say anything general and understandable about how the DB is a concatenation of many individual filings?

docs/glossary.rst
```diff
@@ -0,0 +1,66 @@
...
+^^^^^^
+
+An XBRL filing is how data is actually reported. It contains a list of
+Facts, and a list of Contexts. A filing is associated with a Taxonomy that is used
+to interpret facts and their relationships to one another.
+
+Link Role
```
zaneselvans, line 45: Can we give an example of how this hierarchical arrangement shows up in the FERC Form 1 data, and how it's translated into collections of tables? It almost sounds like Link Roles end up defining foreign key relationships in the DB.  Could we use the income statements, or utility plant balances, which are spread across multiple tables and organized hierarchically into many categories and sub-categories, where one kind of record in one table actually represents an aggregation of all records in another table.

src/ferc_xbrl_extractor/taxonomy.py
```diff
@@ -116,8 +119,36 @@ def from_list(cls, concept_list: list, concept_dict: ConceptDict) -> "Concept":
...
             ],
+            metadata=Metadata.from_concept(concept),
         )
 
+    def get_metadata(
+        self, period_type: Literal["duration", "instant"]
+    ) -> dict[str, dict[str, Any]]:
+        """Get metadata from all leaf nodes in Concept tree.
+
+        Leaf nodes in the Concept tree will become columns in output tables. This method
```
zaneselvans, line 130: If you use the `glossary::` directive in the docs, I think we can also link to the definitions there in the docstrings using e.g. :term:\`concept\`

README.rst
zaneselvans: If it's possible to refer directly to a taxonomy URL on the command line I think we should show an example of that, probably pointing to the same FERC 1 taxonomy URL that it would use by default in the example.

zaneselvans (changes requested):
These docs and examples are very helpful. There are some parts that were still unclear which I've tried to highlight, especially linking the XBRL terminology to the extracted data structures which are being generated by this package (and ultimately distributed & used in PUDL).

README.rst
zschira: I've changed the wording to hopefully make this more clear. If the `path_to_filings` option points to a zipfile, the tool will by default only look for filings there. Technically the filings and the taxonomy can be contained in the same zipfile (as is the case for the archives used by PUDL), but you would still have to indicate that the taxonomy is contained in this zipfile using the `--taxonomy` option, and point to the taxonomy entry point with the `--archive-path` option. This is because a taxonomy is made up of many different files, and the tool doesn't know which of these is the entry point.

As I've added more functionality to the tool (like the ability to parse taxonomies from a zipfile) the CLI has gotten pretty clunky because I was mostly focused on enabling PUDL's usage, but I'd like to rework/streamline it at some point. I think I could allow it to automatically discover the taxonomy entry point, but I just haven't dedicated the time yet.

README.rst
zschira: Added output path

README.rst
zschira: Good call, I've updated all the long commands to be multi-line

README.rst
zschira: Right now it is specifically the 2022 version. I'm not sure when exactly they will release new versions, so I didn't want to try to always use the current year

docs/extract.rst
zschira: Updated to link to tabular datapackage

docs/extract.rst
zschira: Thanks! I tried to do linking, but I wasn't aware of the glossary directive that makes this much easier

docs/glossary.rst
zschira: The relationship is a bit hard to describe, because _some_ Concepts will become columns, but not all of them. I think the figure in the parsing section hopefully illustrates this better. I also added a note about data types, because Concepts have a data type associated with them, that are used to deduce the data types used in the database

docs/glossary.rst
zschira: Added explanation about what the `Context` corresponds to in DB

docs/glossary.rst
zschira: Added more detail, the diagram in the parsing section also shows this relationship

docs/glossary.rst
zschira: Added explanation

docs/glossary.rst
zschira: Added note

docs/glossary.rst
zschira: The hierarchical relationship is shown in the diagram in the parsing section. The relationships they define between Concepts groups them into tables, but does not define any foreign key relationship with other tables. Hopefully this is more clear now

src/ferc_xbrl_extractor/taxonomy.py
zschira: Added links

docs/glossary.rst
zschira: Updated

PR #45, opened 2022-12-20
Arelle 2.3.0

zaneselvans:
* Update to using the most recent version of Arelle (v2.3.0), which avoids the local environment variable contamination
* Remove our context manager workaround for that envvar problem.
* Add the `tox-pytest` workflow so we can make sure we've got whatever tests *do* exist running automatically.
* Revise python versions to just include 3.10 + 3.11 since the CI revealed we had some incompatibilities with older versions.
* Add links to raw + processed data in the README.
* Add some repository status badges in the README.


zaneselvans:
I added the stock `tox-pytest` workflow from `cheshire`. Currently mypy is failing, but I don't know if we expect it to pass. It was complaining about `arelle` imports. I added Arelle to what I thought was the list of imports to ignore in `.mypy.ini` but maybe I don't understand how that works.

@zschira if you think we expect the type checks to succeed and it's an easy tweak to get the configuration set up correctly for that to work, great!  If not we can just remove the `mypy` checks for now.

zschira (approved):
Looks good to me. I've added one small test that will run the extraction using the example filings/taxonomy and verify that it does not produce any errors.

PR #55, opened 2023-02-21
Pandas 2.0

zaneselvans:
Pandas 2.0.0 is out for real, and I'm trying to line everything up in the main PUDL repo to work on it. Since PUDL depends on the XBRL extractor, we need this repo working with pandas 2.0 first.

Thankfully, it seems to work out of the box!

I've updated the dependencies here, and moved to using `pyproject.toml` for all the packaging and project metadata. Tests seem to be passing, so I think this should be okay for a new release here, which we can then pin PUDL to using.


pyproject.toml
```diff
@@ -1,17 +1,109 @@
 [build-system]
+build-backend = "setuptools.build_meta"
 requires = [
-    "setuptools<66",
+    "setuptools>=66,<68",
+    "setuptools_scm[toml]>=3.5.0",
     "wheel",
-    "setuptools_scm[tomp]>=3.5.0",
```
zaneselvans, line 5: typooooo.

pyproject.toml
```diff
@@ -1,17 +1,109 @@
...
+description = "A tool for extracting data from FERC XBRL Filings."
+readme = {file = "README.rst", content-type = "text/x-rst"}
+authors = [
+    {name = "Catalyst Cooperative", email = "pudl@catalyst.coop"},
+    {name = "Zach Schira", email = "zach.schira@catalyst.coop"}
+]
+requires-python = ">=3.10,<3.12"
+dynamic = ["version"]
+license = {file = "LICENSE.txt"}
+dependencies = [
```
zaneselvans, line 20: A bunch of dependencies had stale upper bounds, so I increased them to match what we've got on PUDL. We should get the dependabot working on this repo, otherwise it'll end up constraining what versions PUDL can use, since PUDL depends on this repo.

pyproject.toml
```diff
@@ -1,17 +1,109 @@
...
+include-package-data = true
+
+[tool.setuptools.packages.find]
+where = ["src"]
+
 [tool.setuptools_scm]
 
 [tool.black]
 line-length = 88
-target-version = ["py38", "py39", "py310"]
```
zaneselvans, line 12: These versions were stale so I updated them.

tox.ini
```diff
@@ -141,9 +141,11 @@ commands =
 description = Prepare Python source and binary packages for release.
 basepython = python3
 skip_install = false
+extras =
+    dev
 commands =
     bash -c 'rm -rf build/* dist/*'
-    python setup.py sdist bdist_wheel
+    python -m build
```
zaneselvans, line 148: This is the new basic build backend that replaced running `setup.py`. It creates a source + wheel distribution by default.

pyproject.toml
zschira: I'll see if I can get dependabot working

zschira (approved):
Glad it seems to be working out of the box! I'll also try run the XBRL extraction through PUDL with this version

zaneselvans:
I went ahead and set up dependabot in the repo -- it's good! It'll need to re-run on the `pyproject.toml` file once we merge this all into `main` but that's fine.

PR #106, opened 2023-07-27
Use a test to identify lost facts & assert some stuff about them

### LINKED ISSUE #105: Identify lost XBRL facts
In @zschira 's investigations into https://github.com/catalyst-cooperative/pudl/issues/2755 , he found that there are many XBRL facts that refer to contexts which are missing dimensions, which means that we don't assign those facts to any tables.

We would like to assign those facts to the best of our ability, which means understanding these "lost facts" that don't have a home.

First we should identify them. We may have to implement the notion of a nullable dimension to the table schemas.

jdangerx:
One day, we will be able to run this test in CI and it will tell us when there are too many lost facts.

For now, we're just using the test as a simple script runner that helps us explore the lost facts.

TODO:
* [ ] get more statistics out (e.g. including the context dimensions in the statistics)
* [ ] assert across all instances

We might want to assert something like "in every instance, the % of lost facts named `order_number` is less than 0.5% of the total facts".

Or, we might want to assert something like "in every instance, the % of lost facts at all is less than 1% of the total facts."

Or, "across all instances at once, the % of lost facts is < X%"

### DIFF
--- a/examples/lost_fact_exploration.ipynb
+++ b/examples/lost_fact_exploration.ipynb
@@ -1,0 +1,470 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "a6e96483-25c6-488c-a423-558a0f10cfe1",
+   "metadata": {},
+   "outputs": [],
+   "source": [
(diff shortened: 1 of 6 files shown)


jdangerx:
Closing in favor of catalyst-cooperative/pudl#2774

PR #113, opened 2023-08-09
Lost facts fix

### LINKED ISSUE #2755: FERC Form 1 Statement of Income Contains Holes
### Describe the bug
Upon obtaining the .csv FERC Form 1 files derived from XBRL data using PUDL, we find that some fields have missing data that should not be missing. For example, the file "statement_of_income_114_duration.csv" has missing data in the field "net_interest_charges," but when we check the PDF FERC Form 1 document, the field is not missing. We do not know the full extent of this issue, but has impacted our work.

### Bug Severity
How badly is this bug affecting you?
 -High/Medium: we basically can't use the data from PUDL if we don't trust that it's accurate. Our alternative right now is manually entering data from PDF files for data since the XBRL shift (2021 and 2022).

### To Reproduce
I don't need to provide code--it's simply an issue inside the CSV files in PUDL.

### Expected behavior
I expected there to be data in the field "net_interest_income," but there is not.

### Software Environment?
-Windows

### Additional context
(none).

zschira:
This PR adds an initial fix for FERC missing facts initially identified in [this](https://github.com/catalyst-cooperative/pudl/issues/2755) issue. The notebook in `examples/lost_fact_exploration.ipynb` demonstrates exploration of the problem, and outlines which missing facts are dealt with in this PR. There are still more cases that are identified in the notebook, but need more manual exploration before applying a fix. I will break out a separate issue for tracking the remaining missing facts.


src/ferc_xbrl_extractor/settings.py
```diff
@@ -0,0 +1,8 @@
+"""Define settings to use for extractor."""
+
+FERC1_OPTIONAL_COLUMNS = {
```
zschira, line 3: Not sure if these should be constants, or configurable. And if we want to keep them as constants, I'm open to suggestions on where to place them.

src/ferc_xbrl_extractor/datapackage.py
```diff
@@ -336,7 +337,11 @@ def __init__(self, schema: Schema, period_type: str):
         self.columns = {
             field.name: CONVERT_DTYPES[field.type_] for field in schema.fields
         }
-        self.axes = [name for name in schema.primary_key if name.endswith("axis")]
+        self.axes = {
```
jdangerx, line 340: We had been talking about trying to put the nullability data in the schema, but not wanting to make `fields: list[tuple[Field, bool]]` since that goes against the Frictionless definition of a datapackage.

Anyways. I found that there's this [`required` constraint](https://specs.frictionlessdata.io/table-schema/#constraints) on a `Field` which we could implement, and that would get this nullability information into the right spot.

Separately, what's the relationship between our datapackage classes and the official frictionless-py implementation of their data structures? I see that they support the constraints on a Field already: https://github.com/frictionlessdata/frictionless-py/blob/main/frictionless/schema/field.py#L22

src/ferc_xbrl_extractor/instance.py
```diff
@@ -172,6 +174,9 @@ def as_primary_key(self, filing_name: str) -> dict[str, str]:
                 "end_date": self.period.end_date,
             }
 
+        if optional_axis:
```
jdangerx, line 177: I'm wondering if we should just leave these as null values as we put them into the dataframe. That would mean that, downstream, we'd have to interpret them somehow, but I think in our "convert XBRL to SQLite with minimal interpretation and processing" step we can avoid that interpretation step.

And then we don't have to have something in this very general "FactTable to DataFrame" function that understands the interpretation of every single nullable field in every table context.

src/ferc_xbrl_extractor/settings.py
jdangerx: Each column might be nullable in some FactTables, but not others, right? It might be good to keep track of that, too - and then we can use that information when we're trying to call `Field.from_concept`.

src/ferc_xbrl_extractor/instance.py
jdangerx: If we do want to interpret these values here, we should dynamically look up what the default value is, as defined in `FERC1_OPTIONAL_COLUMNS`.

src/ferc_xbrl_extractor/settings.py
jdangerx: If we just want to set nullability by column name & not on a `(table, column)` basis, this location seems fine for me!

jdangerx (changes requested):
This nullability handling is definitely more flexible than before, but I think we can still get the code to better reflect our conceptual understanding of the correspondence between an XBRL schema and a Frictionless schema.

jdangerx:
Closed in favor of #118 .

PR #117, opened 2023-08-16
Basic FERC data extraction metrics

jdangerx:
After a surprisingly long development cycle, which I'll unpack later, we have some tests that run on a subset of FERC data that spans multiple forms & multiple years.

I decided to sample the data because I found running these metrics across all the FERC XBRL data we have access to, to take a really long time (>1h) and because I didn't want to have to set up Git LFS to Store all these Large Files. Happy to test against everything, but see the note about empty file handling at the bottom.

They measure:
* lost fact %
* non-null values exist in every row
* primary key uniqueness

During development, I refactored some aspects of `extract()` to be easier to test. I haven't looked into downstream impacts on PUDL yet.

I also found some issues with empty file handling, and might have to refactor further to properly handle those, but that's out of scope of this PR.


jdangerx:
Ahh, realizing I didn't run `pre-commit install` for this repo. Which is why none of my pre-commit checks were running before 😅

tests/integration/console_scripts_test.py
jdangerx, line 1: These changes are mostly to fix the usage of pkg_resources and to clean up after ourselves in test.

tests/integration/datapackage_test.py
```diff
@@ -76,8 +76,8 @@ def _create_schema(instant=True, axes=None):
             pd.read_csv(
                 io.StringIO(
                     "c_id,entity_id,filing_name,start_date,end_date,dimension_one_axis,column_one,column_two\n"
-                    'cid_1,EID1,filing,2021-01-01,2021-12-31,Total,"value 1","value 2"\n'
-                    'cid_4,EID1,filing,2020-01-01,2020-12-31,Total,"value 3","value 4"\n'
+                    'cid_1,EID1,filing,2021-01-01,2021-12-31,total,"value 1","value 2"\n'
```
jdangerx, line 79: These should fix the tests that are failing on `main` right now.

src/ferc_xbrl_extractor/cli.py
```diff
@@ -183,22 +183,32 @@ def main():
...
-    instances = get_instances(
-        Path(args.instance_path),
-    )
-
-    xbrl.extract(
-        instances,
-        engine,
-        taxonomy,
-        args.form_number,
+    tables = xbrl.get_fact_tables(
```
jdangerx, line 186: Pulling taxonomy generation out of the main `extract` function makes the interface simpler, but also lets us not double-generate the taxonomies.

jdangerx:
I think @zschira is still the right one to review, but FYI @zaneselvans if you were curious about the test issues.

zschira (changes requested):
Looks good! One small nit about return types, and I think we should make sure to get the logging for duplicate facts implemented here somewhere, but otherwise I'm good to merge

tests/integration/datapackage_test.py
jdangerx: I think it was probably my diff that kept the capital T in when I reformatted the test data.

jdangerx:
I just added the logs & fixed the type signature! I think we have two options here:

1. merge this (once review passes), knowing that there are still some issues with empty files & deep deduplication; then, get #124 in as a follow-up
2. close this and then review #124 as a PR against `main`

I have a light (3/10) preference for keeping these PRs split, but @zschira you're the reviewer - let me know if you have a stronger preference!

src/ferc_xbrl_extractor/instance.py
```diff
@@ -240,6 +254,14 @@ def __init__(
         # This is a nested dictionary of dictionaries to locate facts by context
         self.instant_facts = instant_facts
         self.duration_facts = duration_facts
+        self.all_fact_ids = {
```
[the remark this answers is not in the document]
jdangerx: Good thought! After adding this, I'm finding that this is logging quite a few duplicate facts for some forms. But the deduplication, even the simple version in this PR, seems to handle them nicely :)

zschira (approved):

PR #118, opened 2023-08-17
Total axes

### LINKED ISSUE #2755: FERC Form 1 Statement of Income Contains Holes

zschira:
This PR changes the extractor behavior so missing `Axes` will be treated as Totals.


src/ferc_xbrl_extractor/datapackage.py
```diff
@@ -346,44 +351,34 @@ def construct_dataframe(self, instance: Instance) -> pd.DataFrame:
...
-                    try:
-                        df[key][i] = self.columns[key](val)
-                    except ValueError:
-                        self.logger.warning(
-                            f"Could not parse {val} as {self.columns[key]} in column {key}"
-                        )
-
-        # Create dataframe and drop empty rows
-        return pd.DataFrame(df).dropna(how="all")
+        df = {}
```
jdangerx, line 356: This whole "make a dictionary, fill the columns with a nested for-loop, and then turn that dictionary into a dataframe" thing seemed like a lot of steps, for the goal of "make a dataframe with a bunch of facts."

src/ferc_xbrl_extractor/datapackage.py
```diff
@@ -346,44 +351,34 @@ def construct_dataframe(self, instance: Instance) -> pd.DataFrame:
...
+            {
+                c_id: instance.contexts[c_id].as_primary_key(
+                    instance.filing_name, self.axes
+                )
+                for c_id in df.index
+            }
+        )
+
+        # Join on context ID
+        df = primary_key.T.join(df)
```
jdangerx, line 379: If you want to make this dataframe without having to `T`ranspose it, you can do `pd.DataFrame.from_dict({...}, orient="index")`.

src/ferc_xbrl_extractor/datapackage.py
jdangerx: I do wonder, though, if we want to keep this dataframe stacked/narrow. The unstacked version can get pretty sparse for a lot of forms where each context has one fact for a different concept.

The stacked version would also potentially be much more robust to future taxonomy changes - so long as the primary key is the same / only added to, the concept names can fluctuate and we can just store them all in the same table!

I think that change would definitely be out of scope of this PR.

src/ferc_xbrl_extractor/instance.py
```diff
@@ -251,45 +231,36 @@ def __init__(
...
+            period_fact_dict = self.instant_facts
         else:
-            facts = self.duration_facts.get(tuple(sorted(axes)), {})
+            period_fact_dict = self.duration_facts
 
-        return facts
+        return {
+            fact.c_id: fact
+            for fact in period_fact_dict.get(concept_name, [])
+            if self.contexts[fact.c_id].check_dimensions(primary_key)
```
jdangerx, line 262: If we are joining this to the filtered contexts later, we might not need to filter by "is the context applicable to the table?" here.

jdangerx (changes requested):
The only *actually* blocking thing is the question about whether `dropna()` is doing what we want it to do. But I would feel warm and fuzzy inside if you engaged with my various suggestions, too!

Broader questions I wanted to reiterate at the top level:
* do we need to filter by context dimensions in `get_facts`? it might make sense to just grab *all* the facts. then we can filter `contexts` by whether or not each context fits with the `schema.primary_key`, and avoid passing extra state around.
* should we consider *not* smearing the concept names into their own columns? I'm sure there was some sort of decision made about this in the ancient times, but what was the rationale there?
* should we have the output dataframes be indexed by `schema.primary_key`? 

Finally, when I was testing the code I'm suggesting, I found a couple weird things that I needed to handle:
* empty files, obviously
* filings with duplicated ("c_id", "name") values in facts... that seemed surprising.

jdangerx (approved):

PR #124, opened 2023-08-22
Handle empty files and attempt to deduplicate more deeply

jdangerx:
This is to handle empty files in some filings - there's a bit of a yak-shave here:

* to handle empty files, we filter them out after trying to parse - I chose to pass back `None` as an error value, but we could also raise an error, or filter ahead of time, but whatever we choose...
* we need to run the filtering both in the `data_quality` test and the CLI, so...
* we should unify the "get a taxonomy, read instances, wire them into the actual dataframe-construction" logic...
* so we might as well clean up the CLI code slightly

Unfortunately I added a bit more deduplication logic in here, since I found a deduplication bug when testing this out. I can pull that into #117 if that's easier for you, @zschira .

### DIFF
--- a/src/ferc_xbrl_extractor/arelle_interface.py
+++ b/src/ferc_xbrl_extractor/arelle_interface.py
@@ -11,7 +11,7 @@ from pydantic import BaseModel
 
 
 def _taxonomy_view(taxonomy_source: str | FileSource.FileSource):
-    """Actually use Arelle to get a taxonomy its relationships."""
+    """Actually use Arelle to get a taxonomy and its relationships."""
     cntlr = Cntlr.Cntlr()
     cntlr.startLogging(logFileName="logToPrint")
     model_manager = ModelManager.initialize(cntlr)
--- a/src/ferc_xbrl_extractor/arelle_interface.py
+++ b/src/ferc_xbrl_extractor/arelle_interface.py
@@ -1,4 +1,5 @@
 """Abstract away interface to Arelle XBRL Library."""
+from pathlib import Path
 from typing import Literal
 
 import pydantic
@@ -9,44 +10,41 @@ from arelle.ViewFileRelationshipSet import ViewRelationshipSet
 from pydantic import BaseModel
 
 
-def load_xbrl(path: str | FileSource.FileSource):
-    """Load XBRL (either taxonomy or individual filing).
-
-    Args:
-        path: URL or local path pointing to an XBRL taxonomy or instance.
-    """
+def _taxonomy_view(taxonomy_source: str | FileSource.FileSource):
+    """Actually use Arelle to get a taxonomy its relationships."""
     cntlr = Cntlr.Cntlr()
     cntlr.startLogging(logFileName="logToPrint")
     model_manager = ModelManager.initialize(cntlr)
-    return ModelXbrl.load(model_manager, path)
+    taxonomy = ModelXbrl.load(model_manager, taxonomy_source)
 
-
-def load_taxonomy(path: str | FileSource.FileSource):
-    """Load XBRL taxonomy, and parse relationships.
-
-    Args:
-        path: URL or local path pointing to an XBRL taxonomy.
-    """
-    taxonomy = load_xbrl(path)
-
-    # Interpret structure/relationships
     view = ViewRelationshipSet(taxonomy, "taxonomy.json", "roles", None, None, None)
     view.view(XbrlConst.parentChild, None, None, None)
 
     return taxonomy, view
 
 
-def load_taxonomy_from_archive(filepath: str, archive_path: str):
+def load_taxonomy(path: Path):
+    """Load XBRL taxonomy, and parse relationships.
+
+    Args:
+        path: URL or local path pointing to an XBRL taxonomy.
+    """
+    # arelle only works with `str`, not `Path` - as of version 2.12.2
+    source = str(path)
+    return _taxonomy_view(source)
+
+
+def load_taxonomy_from_archive(filepath: Path, archive_path: Path):
     """Load an XBRL taxonomy from a zipfile archive.
 
     Args:
         filepath: Path to zipfile on disc.
         archive_path: Relative path to taxonomy entry point within archive.
     """
-    # Create arelle FileSource object
-    f = FileSource.openFileSource(archive_path, sourceZipStream=filepath)
-
-    return load_taxonomy(f)
+    file_source = FileSource.openFileSource(
+        str(archive_path), sourceZipStream=str(filepath)
+    )
+    return _taxonomy_view(file_source)
 
 
 class References(BaseModel):
--- a/src/ferc_xbrl_extractor/cli.py
+++ b/src/ferc_xbrl_extractor/cli.py
@@ -119,8 +119,8 @@ def run_main(
         file_logger.setFormatter(logging.Formatter(log_format))
         logger.addHandler(file_logger)
 
-    db_path = f"sqlite:///{sql_path}"
-    engine = create_engine(db_path)
+    db_uri = f"sqlite:///{sql_path}"
+    engine = create_engine(db_uri)
 
     if clobber:
         helpers.drop_tables(engine)
@@ -141,7 +141,7 @@ def run_main(
     extracted = xbrl.extract(
         taxonomy_path=taxonomy,
         form_number=form_number,
-        db_path=db_path,
+        db_uri=db_uri,
         archive_path=archive_path,
         datapackage_path=datapackage_path,
         metadata_path=metadata_path,
@@ -151,10 +151,10 @@ def run_main(
     )
 
     with engine.begin() as conn:
-        for table_name, filing in extracted.filings.items():
+        for table_name, data in extracted.table_data.items():
             # Loop through tables and write to database
-            if not filing.empty:
-                filing.to_sql(table_name, conn, if_exists="append")
+            if not data.empty:
+                data.to_sql(table_name, conn, if_exists="append")
 
 
 def main():
--- a/src/ferc_xbrl_extractor/cli.py
+++ b/src/ferc_xbrl_extractor/cli.py
@@ -1,9 +1,7 @@
 """A command line interface (CLI) to the xbrl extractor."""
 import argparse
-import io
 import logging
 import sys
-import zipfile
 from pathlib import Path
 
 import coloredlogs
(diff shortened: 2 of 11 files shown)


src/ferc_xbrl_extractor/arelle_interface.py
```diff
@@ -9,44 +10,41 @@
...
+    taxonomy = ModelXbrl.load(model_manager, taxonomy_source)
+
+    view = ViewRelationshipSet(taxonomy, "taxonomy.json", "roles", None, None, None)
+    view.view(XbrlConst.parentChild, None, None, None)
+
+    return taxonomy, view
 
 
-def load_taxonomy(path: str | FileSource.FileSource):
+def load_taxonomy(path: Path):
```
jdangerx, line 26: I wanted to use `Paths` in as much of our code as possible so I pushed the `str` conversion down to "right before we interface with Arelle."

src/ferc_xbrl_extractor/cli.py
```diff
@@ -34,8 +31,9 @@ def parse_main():
     )
     parser.add_argument(
         "-s",
-        "--save-datapackage",
+        "--datapackage-path",
```
jdangerx, line 34: This name is more consistent with e.g. `metadata-path`.

src/ferc_xbrl_extractor/cli.py
```diff
@@ -76,139 +74,93 @@ def parse_main():
...
-        for instance in sorted(instances)
-        if instance.suffix in allowable_suffixes
-    ]
-
-
-def main():
-    """CLI for extracting data fro XBRL filings."""
-    args = parse_main()
-
+def run_main(
```
jdangerx, line 97: In general I like having a function that handles all the actual post-parsing logic, so we can call this function from a non-CLI environment if we need to.

src/ferc_xbrl_extractor/cli.py
```diff
@@ -76,139 +74,93 @@ def parse_main():
...
         help="Set log level (valid arguments include DEBUG, INFO, WARNING, ERROR, CRITICAL)",
         default="INFO",
     )
-    parser.add_argument("--logfile", help="Path to logfile", default=None)
+    parser.add_argument("--logfile", help="Path to logfile", type=Path, default=None)
 
     return parser.parse_args()
 
 
-def instances_from_zip(instance_path: Path) -> list[InstanceBuilder]:
```
jdangerx, line 99: Moved these to `instance.py` for now, though I'm not quite sure if that's the right spot. There's a good argument for `xbrl.py` I think.

src/ferc_xbrl_extractor/instance.py
```diff
@@ -321,13 +323,11 @@ def parse(self, fact_prefix: str = "ferc") -> Instance:
         parser = etree.XMLParser(huge_tree=True)
 
         # Check if instance contains path to file or file data and parse accordingly
-        if isinstance(self.file, str):
+        try:
             tree = etree.parse(self.file, parser=parser)  # nosec: B320
-            root = tree.getroot()
-        elif isinstance(self.file, io.BytesIO):
```
jdangerx, line 327: An `io.BytesIO` is "file-like" so we can just shove it into `etree.parse()` regardless.

src/ferc_xbrl_extractor/instance.py
```diff
@@ -321,13 +323,11 @@ def parse(self, fact_prefix: str = "ferc") -> Instance:
...
-            root = etree.fromstring(self.file.read(), parser=parser)  # nosec: B320
-        else:
-            raise TypeError("Can only parse XBRL from path to file or file like object")
+        except etree.XMLSyntaxError:
+            return None
```
jdangerx, line 329: We could move this try/except into `xbrl.extract`... then this function will just raise an exception, which I think is maybe nicer? Future callers get to interpret the exception instead of us swallowing it here.

src/ferc_xbrl_extractor/taxonomy.py
```diff
@@ -230,8 +230,8 @@ class Taxonomy(BaseModel):
     @classmethod
     def from_path(
         cls,
-        path: str,
-        archive_file_path: str | None = None,
+        path: Path,
+        archive_path: Path | None = None,
```
jdangerx, line 234: Rename for consistency with the command-line arg, but one day it'd be nice to do away with this weird double-path thing.

src/ferc_xbrl_extractor/xbrl.py
```diff
@@ -12,11 +13,53 @@
...
-from ferc_xbrl_extractor.instance import Instance
+from ferc_xbrl_extractor.instance import Instance, get_instances
 from ferc_xbrl_extractor.taxonomy import Taxonomy
 
+ExtractOutput = namedtuple("ExtractOutput", ["tables", "instances", "filings", "stats"])
 
+
+# TODO (daz): make taxonomy_path take a union type of "taxonomy location" = URL | (Path, Path) and then ditch archive_path arg
 def extract(
+    taxonomy_path,
```
jdangerx, line 24: TODO: these need types

tests/integration/data_quality_test.py
```diff
@@ -33,21 +30,16 @@ def data_dir(request) -> Path:
...
-        batch_size=max(1, len(instances) // 5),
-    )
-    return ExtractOutput(
-        tables=tables, instances=instances, filings=filings, stats=stats
+        archive_path=None,
+        metadata_path=metadata_dir / "metadata.json",
+        datapackage_path=metadata_dir / "datapackage.json",
+        instance_path=data_dir / f"ferc{form}-xbrl-{year}.zip",
+        workers=None,
+        batch_size=2,
```
jdangerx, line 42: TODO: maybe we refactor this to `num-batches`?

jdangerx (commented):
Something halfway between a collection of notes to self and notes to you, dear reviewer.

tests/integration/data_quality_test.py
jdangerx: I guess `batch_size` is better for CLI (you don't know how many files there are) but `num_batches` is more helpful for test (you just care that it gets split up in some way)... I'll eat the slightly more annoying testing.

src/ferc_xbrl_extractor/instance.py
jdangerx: Tried pulling this out, and found the original place where this handling was occuring in `xbrl.py` so I think it's the right move.

src/ferc_xbrl_extractor/datapackage.py
```diff
@@ -400,34 +398,72 @@ class Datapackage(BaseModel):
...
+
+    We pull out the duplicated values, group them by index, and then see if
+    the values are "close enough" to each other. If they are, we pull in the
+    first value, and combine all the values together back into one dataframe.
+
+    The relative tolerance is set to 2e-4 because there is one data point in
+    FERC Form 1 in 2021/2022 outside of the `np.isclose()` default 1e-5
+    tolerance:
+
+    For one context, concept `length_for_stand_alone_transmission_lines`
```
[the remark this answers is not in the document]
jdangerx: Update - this appears to only be the concepts "length_for_stand_alone_transmission_lines" and "length_for_transmission_lines_aggregated_with_other_structures", and sometimes we get a rounded value alongside a non-rounded value.

What I'm seeing in the 2022 data is:

```
41695:  <ferc:LengthForStandAloneTransmissionLines id="ad3VybDovL2RvY3MudjEvZG9jOjNjMGRhYjdkNWU3YjQ5ZWE5ZmNjMTQzZDdmNzIxYjlmL3RhYmxlOjJjZWFiZjE2YTk5NTRiZWM4MzU2ZGVmMTdkNjlmZGEyL3RhYmxlY2VsbDoyY2VhYmYxNmE5OTU0YmVjODM1NmRlZjE3ZDY5ZmRhMjAwMDAwMDAwMDkwMDAwMDAwMDQ2OTIyMzM3MjAzNjg1NDcwODExMw_cell-74-J47-36-f" contextRef="if210855906b944d39585c4e18cd5a0a8_D20220101-20221231" decimals="2" unitRef="id97a4ac198c44e4eb49bfa14db95c6cc">519.82</ferc:LengthForStandAloneTransmissionLines>
42123:  <ferc:LengthForStandAloneTransmissionLines id="ad3VybDovL2RvY3MudjEvZG9jOjNjMGRhYjdkNWU3YjQ5ZWE5ZmNjMTQzZDdmNzIxYjlmL3RhYmxlOjU0NDNjYWQ1ZTQ5ZjQxYWViY2JkZWE5MmEyOWZlMjgxL3RhYmxlY2VsbDo1NDQzY2FkNWU0OWY0MWFlYmNiZGVhOTJhMjlmZTI4MTAwMDAwMDAwMDkwMDAwMDAwMDQ2OTIyMzM3MjAzNjg1NDcwODExMw_cell-75-J47-36-f" contextRef="if210855906b944d39585c4e18cd5a0a8_D20220101-20221231" decimals="0" unitRef="id97a4ac198c44e4eb49bfa14db95c6cc">520</ferc:LengthForStandAloneTransmissionLines>
```

Where you can see that `decimals="2"` vs. `decimals="0"` for the unrounded/rounded versions of the fact.

You can also see that the fact `id` column is slightly different, but the context ref and concept name are the same. 

I suppose we can track the decimals field too, and defer to the more precise measurement where possible.

src/ferc_xbrl_extractor/instance.py
```diff
@@ -372,3 +369,53 @@ def parse(self, fact_prefix: str = "ferc") -> Instance:
                     duration_facts[new_fact.name].append(new_fact)
 
         return Instance(context_dict, instant_facts, duration_facts, self.name)
+
+
+def instances_from_zip(instance_path: Path) -> list[InstanceBuilder]:
```
jdangerx, line 374: These two functions are just copy-pasted.

src/ferc_xbrl_extractor/taxonomy.py
zschira: We can probably deduce the path to the entry point and not pass this, but probably something we can leave for the future.

zschira (approved):
Looks great to me!

src/ferc_xbrl_extractor/datapackage.py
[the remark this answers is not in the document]
jdangerx: Rather than refactor to take the decimals field into account, I just added a workaround to always take the most precise amount.

PR #142, opened 2023-09-22
Fix memory issues

zschira:
This PR fixes the high memory usage introduced in `1.0.0`. This problem was caused by parsing `Instance`'s at the beginning, and passing the parsed objects to the worker processes. These objects are quite large, and get copied when passed to subprocesses, which lead to huge increases in total memory usage. This PR reverts to parsing these objects when they are needed in each subprocess. It also slightly reworks some integration tests to accommodate this change.


jdangerx (approved):
Wow, this turned out a lot cleaner than I expected - didn't really have to change too much! Nice :)

PR #144, opened 2023-09-24
Pandas performance & automated release

zaneselvans:
* Address performance issues that surfaced under pandas 2.1
* Add an automated release workflow that will attempt to publish a new package when a tag starting with `v` is pushed.
* Address some deprecation warnings in the command line interface test cases.
* Fix some bad URLs in the package metadata.
* Additional keywords in package metadata.

### DIFF
--- a/src/ferc_xbrl_extractor/xbrl.py
+++ b/src/ferc_xbrl_extractor/xbrl.py
@@ -84,9 +84,11 @@ def _dedupe_newer_report_wins(df: pd.DataFrame, primary_key: list[str]) -> pd.Da
     old_index = df.index.names
     return (
         df.reset_index()
+        .sort_values("report_date")
         .groupby(unique_cols)
-        .apply(lambda df: df.sort_values("report_date").ffill().iloc[-1])
+        .last()
         .drop("report_date", axis="columns")
+        .reset_index()
         .set_index(old_index)
     )
 
--- a/tests/integration/console_scripts_test.py
+++ b/tests/integration/console_scripts_test.py
@@ -19,7 +19,7 @@ def test_extractor_scripts(script_runner, ep):
 
     The script_runner fixture is provided by the pytest-console-scripts plugin.
     """
-    ret = script_runner.run(ep, "--help", print_result=False)
+    ret = script_runner.run([ep, "--help"], print_result=False)
     assert ret.success  # nosec: B101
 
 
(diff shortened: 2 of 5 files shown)


src/ferc_xbrl_extractor/xbrl.py
```diff
@@ -84,9 +84,11 @@ def _dedupe_newer_report_wins(df: pd.DataFrame, primary_key: list[str]) -> pd.Da
     old_index = df.index.names
     return (
         df.reset_index()
+        .sort_values("report_date")
```
zaneselvans, line 87: We can `sort_values()` before doing the groupby because according to the [groupby docs](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.groupby.html) "Groupby preserves the order of rows within each group."

src/ferc_xbrl_extractor/xbrl.py
```diff
@@ -84,9 +84,11 @@ def _dedupe_newer_report_wins(df: pd.DataFrame, primary_key: list[str]) -> pd.Da
...
         .groupby(unique_cols)
-        .apply(lambda df: df.sort_values("report_date").ffill().iloc[-1])
+        .last()
```
zaneselvans, line 89: `groupby.last()` [selects the last non-null value from each column](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.core.groupby.DataFrameGroupBy.last.html).

tox.ini
```diff
@@ -109,37 +108,3 @@ commands =
...
-[testenv:release]
-description = Release the package to the production PyPI server.
-basepython = python3
-skip_install = false
-extras =
-    dev
-commands =
-    {[testenv:build]commands}
-    twine check dist/*
-    twine upload --verbose --skip-existing dist/*
```
zaneselvans, line 145: Removed because it takes place in the `release` workflow now.

tests/integration/console_scripts_test.py
```diff
@@ -19,7 +19,7 @@ def test_extractor_scripts(script_runner, ep):
 
     The script_runner fixture is provided by the pytest-console-scripts plugin.
     """
-    ret = script_runner.run(ep, "--help", print_result=False)
+    ret = script_runner.run([ep, "--help"], print_result=False)
```
zaneselvans, line 22: Just a switch from deprecated syntax.

src/ferc_xbrl_extractor/xbrl.py
zaneselvans: One potentially non-deterministic outcome here is that if there are duplicate values of `report_date` within the column then which set of data ends up being "last" might not always be the same. Do we think there can ever be multiple filings within a group that have the same date?

jdangerx: Yes! 

For some reason, I made this issue in the PUDL repo instead: https://github.com/catalyst-cooperative/pudl/issues/2822 but the short of it is "we need to use metadata from `rssfeed` to sort them better."

src/ferc_xbrl_extractor/xbrl.py
jdangerx: Perfect! That's the exact intent & also this conveys what's going on way better!

tox.ini
jdangerx: wooohoo!

PR #147, opened 2023-09-27
Force tables to have all columns that are defined in schema

jdangerx:
In https://github.com/catalyst-cooperative/pudl/issues/2897 I found that we were missing some columns because the `.unstack()` in `construct_dataframe` doesn't create columns for values that don't show up at all, even if they're defined in the metadata. Applying a reindex makes sure we get everything.

This also was causing some integration test failures - when running the ETL in-process, we would:

1. write 2021 data for a table
2. construct dataframe for 2022 data, which has a slightly different column set because of different reported values
3. fail when trying to write 2022 data to SQLite

Lastly, I wonder if there's a way we could keep our extracted tables tidy - our transforms in PUDL promptly re-stack these wide tables in `wide_to_tidy`, so maybe we can skip that completely. But that's definitely out of scope of this PR.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,7 +26,7 @@ repos:
   # Formatters: hooks that re-write Python and RST files
   ########################################################################################
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.0.290
+    rev: v0.0.291
     hooks:
       - id: ruff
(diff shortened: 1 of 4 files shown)


jdangerx:
Yeah, I think we plant our mental seeds now and then reap them when we have to integrate 2023 data 🌱

PR #151, opened 2023-10-06
Track publication time in primary key & stop deduping rows

jdangerx:
For each table, we'll now have a row per context per filing, which includes all the relevant facts reported in that context.

In 0.8.3 and below, we relied on the caller to only request one filing per entity.

In 1.1.1 and below, we read in all the filings for an entity, but then deduplicated the table so that each fact would only be reported once. We did this by sorting by the report date, and then picking the last reported value for each fact.

Unfortunately, since report date is not granular enough, this lead to an ambiguous sort order, which led to some issues matching data between tables since they would be associated with different filing names.

We will stop trying to tinker with the data here, beyond bringing it into SQLite form, and do the deduplication within PUDL. That means we will include duplicate facts whenever there are multiple filings reporting the same fact.


jdangerx:
Nooo the `published_parsed` field is in EDT and the runners run on UTC.

Someday we need to update the archiver to put the `published_parsed` in UTC, but until now we will explicitly use UTC-4.

tests/integration/data_quality_test.py
```diff
@@ -54,11 +54,11 @@ def test_lost_facts_pct(extracted, request):
         # We have unallocated data for Form 6 for some reason.
         total_threshold = 0.9
         per_filing_threshold = 0.8
-        # Assert that this is < 0.95 so we remember to fix this test once we
+        # Assert that this is < 0.96 so we remember to fix this test once we
```
jdangerx, line 57: This number went up a little bit, but not a lot, and we didn't actively try to fix anything here, so I just bumped the "we still need to fix it" threshold.

tests/integration/data_quality_test.py
```diff
@@ -71,17 +71,38 @@ def test_lost_facts_pct(extracted, request):
...
-        assert not df.reset_index().duplicated(unique_cols).any()
+        assert (
+            table_data[table_name]
+            .reset_index(level="publication_time")
+            .publication_time.notna()
+            .all()
+        )
+
+
+def test_all_data_has_corresponding_id(extracted):
```
jdangerx, line 86: @zaneselvans FYI I wrote a test to make sure this change does give us the guarantee we want.

jdangerx:
Nooo the `published_parsed` field is actually in Eastern Time, which may or may not be Eastern Daylight Time. Updating to force `America/New_York`.

src/ferc_xbrl_extractor/instance.py
```diff
@@ -390,16 +408,52 @@ def instances_from_zip(instance_path: Path | io.BytesIO) -> list[InstanceBuilder
 
     archive = zipfile.ZipFile(instance_path)
 
+    with archive.open("rssfeed") as f:
```
zschira, line 411: Out of scope, but this is a place I think it would be really nice to have access to the `Datastore`. Given that we're basically requiring our specific archive layout for the input data, it would be nice to have a single unified interface to the archives.

zschira (approved):
Looks good!

PR #169, opened 2023-11-25
Update to Pydantic v2 and Ruff formatting

zaneselvans:
* Minor changes to the code to use Pydantic v2. No tricky refactoring, all tests passed.
* Switched from using `black` to `ruff` for autoformatting. Did not result in any changes.
* Addressed or suppressed a variety of deprecation warnings that have been blowing up the logs in PUDL.


jdangerx (approved):
Thanks for the upgrade! Glad it was straightforward.

PR #187, opened 2024-01-20
Update dependencies to allow pandas 2.2 and improve performance

zaneselvans:
- Allow use of pandas 2.2
- Require pyarrow, which will be required by pandas 3.0, and now raises a deprecation warning. Should also improve string performance in some use cases.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -4,9 +4,9 @@ channels:
   - defaults
 dependencies:
   # Packages required for setting up the environment
-  - pip>=21.0,<24
+  - pip>=21.0
   - python>=3.10,<3.13
-  - setuptools>=66,<69
+  - setuptools>=66
 
   # XBRL parsing library
   - pip:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
 build-backend = "setuptools.build_meta"
-requires = ["setuptools>=66,<69", "setuptools_scm[toml]>=3.5.0", "wheel"]
+requires = ["setuptools>=66", "setuptools_scm[toml]>=3.5.0", "wheel"]
 
 [project]
 name = "catalystcoop.ferc_xbrl_extractor"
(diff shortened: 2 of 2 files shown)


pyproject.toml
zaneselvans, line 1: Most of the changes in here are because my TOML linter went to town.

environment.yml
```diff
@@ -4,9 +4,9 @@ channels:
   - defaults
 dependencies:
   # Packages required for setting up the environment
-  - pip>=21.0,<24
+  - pip>=21.0
   - python>=3.10,<3.13
-  - setuptools>=66,<69
+  - setuptools>=66
```
zaneselvans, line 9: Unnecessarily restrictive.

pyproject.toml
```diff
@@ -1,32 +1,29 @@
 [build-system]
 build-backend = "setuptools.build_meta"
-requires = [
-    "setuptools>=66,<69",
-    "setuptools_scm[toml]>=3.5.0",
-    "wheel",
-]
+requires = ["setuptools>=66", "setuptools_scm[toml]>=3.5.0", "wheel"]
```
zaneselvans, line 3: Allow most recent version of setuptools.

pyproject.toml
```diff
@@ -1,32 +1,29 @@
...
+    "lxml>=4.9.1,<6",
+    "numpy>=1.16,<2",
+    "pandas>=1.5,<3",
+    "pyarrow>=14.0.1", # required starting in pandas 3.0
+    "pydantic>=2,<3",
     "sqlalchemy>=1.4,<3",
-    "pandas>=1.5,<2.2",
     "stringcase>=1.2,<2",
-    "numpy>=1.16,<2",
-    "lxml>=4.9.1,<6",
```
zaneselvans, line 29: Only real changes here are allowing pandas 2.2 and requiring pyarrow.

PR #205, opened 2024-03-22
Add retries to taxonomy reading.

jdangerx:
# Overview

Addresses catalyst-cooperative/pudl#3449 

When we try to load a taxonomy multiple times concurrently, we run into `FileExistsError`s as the various threads try to cache files at the same location.

I made us retry the taxonomy read if the cache is confused.

I also added a surprisingly hard-to-write test that runs these taxonomy reads in parallel.

# Testing

How did you make sure this worked? How can a reviewer verify this?

# To-do list

```[tasklist]
- [ ] after merge: need to push a `v1.3.3` tag
- [ ] depend on new version in `pudl`
```

### DIFF
--- a/src/ferc_xbrl_extractor/arelle_interface.py
+++ b/src/ferc_xbrl_extractor/arelle_interface.py
@@ -1,6 +1,7 @@
 """Abstract away interface to Arelle XBRL Library."""
 
 import io
+import time
 from pathlib import Path
 from typing import Literal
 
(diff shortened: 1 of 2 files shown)


jdangerx:
Thanks for the review! I had just tagged @catalyst-cooperative/inframundo for review & the round-robin picked @zschira - I had no strong preference for review here :)

Fortunately, the FERC monthly archives should already be fixed by [this PR](https://github.com/catalyst-cooperative/pudl-archiver/pull/304). But this is part of getting the PUDL nightly builds to be more reliable, so still worth getting it in sooner rather than later.

PR #233, opened 2024-06-25
Ferc xbrl updates

zschira:
## Background

The `pudl-archiver` has been updated to decouple taxonomies from years of XBRL data. Now it puts all versions of taxonomies in a single zipfile, so if filings from the same year use different versions of the taxonomy, those can all be referenced. This PR updates the extractor to accommodate this change. Now, it will parse all taxonomies and create a dictionary which maps these parsed taxonomies to the version. Then, while parsing individual filings, it will detect the version of the taxonomy referenced in the filing, and use that version for interpreting the facts in the filing.


src/ferc_xbrl_extractor/instance.py
```diff
@@ -431,34 +440,13 @@ def instances_from_zip(instance_path: Path | io.BytesIO) -> list[InstanceBuilder
...
             Path(filename).stem,
             publication_time=publication_times[filename],
+            taxonomy_version=taxonomy_versions[filename],
         )
         for filename in archive.namelist()
         if Path(filename).suffix in allowable_suffixes
     ]
 
 
-def get_filing_name(filing_metadata: dict[str, str | int]) -> str:
```
jdangerx, line 440: Woohoo!

src/ferc_xbrl_extractor/xbrl.py
```diff
@@ -231,11 +226,9 @@ def get_fact_tables(
     Data-Package descriptor describing the output database if requested.
 
     Args:
-        taxonomy_path: URL of taxonomy.
+        taxonomy_path: Zipfile with archived taxonomies for form.
```
jdangerx, line 229: nit: This arg is called `taxonomy_source` now.

tests/integration/data_quality_test.py
```diff
@@ -58,7 +52,7 @@ def test_lost_facts_pct(extracted, request):
         # fix the bug. We don't use xfail here because the parametrization is
         # at the *fixture* level, and only the lost facts tests should fail
         # for form 6.
-        assert used_fact_ratio > total_threshold and used_fact_ratio <= 0.96
+        assert used_fact_ratio > total_threshold and used_fact_ratio <= 0.97
```
jdangerx, line 55: nit: comment above still says 0.96

src/ferc_xbrl_extractor/xbrl.py
```diff
@@ -85,7 +83,7 @@ def extract(
 
 def table_data_from_instances(
     instance_builders: list[InstanceBuilder],
-    table_defs: dict[str, FactTable],
+    table_defs: dict[str, dict[str, FactTable]],
```
jdangerx, line 86: We had talked about batching instances such that each batch only had to parse one taxonomy - did you find that reading all the taxonomies for each worker turned out OK in the end?

tests/unit/instance_test.py
```diff
@@ -207,6 +192,7 @@ def test_all_fact_ids():
         instant_facts=instant_facts,
         duration_facts=duration_facts,
         filing_name="test_instance",
+        taxonomy_version="form-1_2022-01-01",
```
jdangerx, line 195: nit: it might be good to differentiate `taxonomy_version` in `Instance`, which doesn't expect `.zip`, from `taxonomy_version` in `InstanceBuilder`, which does.

jdangerx (approved):
Seems fine, thanks for updating the tests too! There's a few small nits you can take or leave.

Don't forget to actually do a versioned release so that PUDL can depend on this -_-

src/ferc_xbrl_extractor/xbrl.py
zschira: Fixed!

tests/integration/data_quality_test.py
zschira: Fixed!

src/ferc_xbrl_extractor/xbrl.py
zschira: Actually I'm just parsing all the taxonomies at the beginning, then pass a dict of those to each worker. It's a bit less memory efficient (but not that bad), but only does the parsing once.

tests/unit/instance_test.py
zschira: There all actually using the `.zip` version, fixed!

Issue #153, opened 2023-10-08
Datapackage descriptors are invalid


zaneselvans:
The datapackage descriptors we are currently generating to annotate the SQLite DBs which are derived from XBRL data are not valid.  For example, in the `ferc-xbrl-extractor` environment running this command:

```bash
frictionless validate ferc714_xbrl_datapackage.json
```

Results in a bunch of errors like:
```
# -------
# invalid: sqlite:////Users/zane/code/catalyst/pudl-work/output/ferc714_xbrl.sqlite
# -------

## Summary

+-----------------------------+--------------------------------------------------------------------------+
| Description                 | Size/Name/Count                                                          |
+=============================+==========================================================================+
| File name (Not Found)       | sqlite:////Users/zane/code/catalyst/pudl-work/output/ferc714_xbrl.sqlite |
+-----------------------------+--------------------------------------------------------------------------+
| File size                   | N/A                                                                      |
+-----------------------------+--------------------------------------------------------------------------+
| Total Time Taken (sec)      | 0.002                                                                    |
+-----------------------------+--------------------------------------------------------------------------+
| Total Errors                | 1                                                                        |
+-----------------------------+--------------------------------------------------------------------------+
| Scheme Error (scheme-error) | 1                                                                        |
+-----------------------------+--------------------------------------------------------------------------+

## Errors

+-------+---------+---------+---------------------------------------------------+
| row   | field   | code    | message                                           |
+=======+=========+=========+===================================================+
|       |         | scheme- | The data source could not be successfully loaded: |
|       |         | error   | cannot create loader "". Try installing           |
|       |         |         | "frictionless-"                                   |
+-------+---------+---------+---------------------------------------------------+
```

Or if we try and validate a single resource and return the errors in JSON form:
```bash
frictionless validate --json --resource-name planning_area_hourly_demand_and_forecast_summer_and_winter_peak_demand_and_annual_net_energy_for_load_table_03_2_instant ferc714_xbrl_datapackage.json
```
```json
{
  "version": "4.40.11",
  "time": 0.001,
  "errors": [],
  "tasks": [
    {
      "resource": {
        "path": "sqlite:////Users/zane/code/catalyst/pudl-work/output/ferc714_xbrl.sqlite",
        "profile": "tabular-data-resource",
        "name": "planning_area_hourly_demand_and_forecast_summer_and_winter_peak_demand_and_annual_net_energy_for_load_table_03_2_instant",
        "dialect": {
[...]

PR #242, opened 2024-07-16
Fix datapackage generation from multiple taxonomies

zschira:
# Overview
This PR fixes datapackage generation when using multiple taxonomies to extract XBRL filings. This fixes a metadata issue in the on-going 2023 FERC integration. It does this by merging multiple XBRL taxonomies into a single datapackage before parsing filings. This datapackage is used to structure the output DB and interpret filings during extraction.


src/ferc_xbrl_extractor/datapackage.py
```diff
@@ -423,14 +454,27 @@ def from_taxonomy(
...
         """
-        resources = []
-        for role in taxonomy.roles:
-            for period_type in ["duration", "instant"]:
-                resource = Resource.from_link_role(role, period_type, db_uri)
-                if resource:
-                    resources.append(resource)
+        resources = {}
+        logger.info("Attempting to merge taxonomies into a single datapackage.")
+        for i, (taxonomy_version, taxonomy) in enumerate(sorted(taxonomies.items())):
```
zschira, line 459: Merging taxonomies is a bit weird, so I'll try to explain the logic a bit. `Resource`s in a datapackage correspond to a single table, so this will parse a single taxonomy normally, then for all subsequent taxonomies, it will look for any new or missing tables. For tables that existed in the previous version and the new version, it will call `merge_resources`, which then looks at the columns as well as the primary key of the table.

src/ferc_xbrl_extractor/datapackage.py
```diff
@@ -332,6 +334,36 @@ def get_period_type(self):
         period_type = "instant" if "date" in self.schema_.primary_key else "duration"
         return period_type
 
+    def merge_resources(self, other: "Resource", other_version: str) -> "Resource":
```
zschira, line 337: This is attempting to merge the schema of the same table from two different versions of a taxonomy. It first looks at primary key and raises an error if it's changed. Right now this never occurs, so it's not a problem, but if future versions change a primary key of a table, we'll need to decide how to handle this. After checking the primary key, it will look for any new or missing columns. If there are new columns, it will append these. Changes in the set of columns will lead to Nulls in the extracted table. So far there haven't been many changes in the column set, fortunately.

zschira (commented):
The XBRL extraction is pretty janky, and this change is kind of adding some more jank on top, so I don't feel the best about it tbh. That being said, I waffled a lot on whether I should try to do something better, but ultimately feel like we just need to get 2023 data out the door, so maybe it's good enough for now.

src/ferc_xbrl_extractor/datapackage.py
zaneselvans: This is similar the actual behavior of the DBF derived data, in which we verified that as the DB schema evolved over the years, it was only ever additive -- new tables, and new columns within existing tables, but never the removal of a table of columns, which meant that we could always just use the latest published database to define a schema which was able to contain all years of data. But here you don't have that temporal dependency.

src/ferc_xbrl_extractor/datapackage.py
```diff
@@ -332,6 +334,36 @@ def get_period_type(self):
...
+        """Merge same resource from multiple taxonomies."""
+        if self.schema_.primary_key != other.schema_.primary_key:
+            raise RuntimeError(
+                f"Can't merge resource {self.name} when versions have incompatible schemas"
+            )
+        original_fields = {field.name for field in self.schema_.fields}
+        other_fields = {field.name for field in other.schema_.fields}
+
+        if missing_fields := original_fields - other_fields:
```
e-belfer, line 346: To clarify: does this mean we'll lose the metadata for older columns that have been dropped in the newer data?

src/ferc_xbrl_extractor/datapackage.py
e-belfer: Can we note this in the docstring of this method? Seems helpful to note what the underlying merge behavior is.

src/ferc_xbrl_extractor/datapackage.py
zschira: Yeah the older columns will remain, they're just logged so we can see what changed. I've added this to the docstring now.

src/ferc_xbrl_extractor/datapackage.py
zschira: @zaneselvans that makes sense, Dazhong and I have talked about making the extractor produce tidy tables, which would allow us to handle these changes in a much cleaner way, but decided to hold off for now.

@e-belfer yes! Sorry forgot to really go back and improve the documentation, but should be better now.

src/ferc_xbrl_extractor/datapackage.py
```diff
@@ -332,6 +334,36 @@ def get_period_type(self):
...
+            logger.warning(
+                f"The following fields were removed from table {self.name}"
```
e-belfer, line 348: You're missing an extra space after `{self.name}` here.

e-belfer (changes requested):
I successfully extracted the 2023 FERC XBRL data with this code change. 

One extremely minor line edit in this PR, which is my only actual blocking concern.

Just want to confirm for my sanity that recent XBRL changes break our ability to retroactively extract data from older Zenodo archives, and that this is an expected and intended decision?

Possibly out-of-scope: Are you expecting the datapackages to be valid frictionless datapackages (a la https://github.com/catalyst-cooperative/ferc-xbrl-extractor/issues/153)? I'm seeing tons of errors when I try to validate the one I generated. Again, just let me know if this one is out of scope.

One bigger concern: I _think_ that this update (or one before, or many before...) breaks the sample extraction code in the README. Which means that our instruction docs are now incorrect. We should definitely fix this. If it's quick to address, I'd prefer to do it here. If not, we should open an issue for this and make it a high priority after the 2023 data is merged.

src/ferc_xbrl_extractor/xbrl.py
```diff
@@ -252,30 +252,24 @@ def get_fact_tables(
...
-                    )
+        if not report.valid:
+            raise RuntimeError(f"Generated datapackage is invalid - {report.errors}")
 
-                # Write to JSON file
-                with Path(datapackage_path).open(mode="w") as f:
-                    f.write(datapackage.model_dump_json(by_alias=True))
+        # Write to JSON file
+        with Path(datapackage_path).open(mode="w") as f:
+            f.write(datapackage.model_dump_json(by_alias=True))
```
e-belfer, line 270: Elsewhere we use `indent=n` to make our datapackages far more readable. Maybe worth considering here as the datapackage is currently hard to read.

e-belfer: See e.g., in our archiver repo:

new_datapackage.model_dump_json(by_alias=True, indent=4),
                    encoding="utf-8",

src/ferc_xbrl_extractor/datapackage.py
zschira: Fixed!

src/ferc_xbrl_extractor/xbrl.py
zschira: Good call, fixed!

zschira:
> Possibly out-of-scope: Are you expecting the datapackages to be valid frictionless datapackages (a la https://github.com/catalyst-cooperative/ferc-xbrl-extractor/issues/153)? I'm seeing tons of errors when I try to validate the one I generated. Again, just let me know if this one is out of scope.

Hmm how were you checking the validation? We have an integration test, and validate the datapackage before writing to a json file.

> One bigger concern: I think that this update (or one before, or many before...) breaks the sample extraction code in the README. Which means that our instruction docs are now incorrect. We should definitely fix this. If it's quick to address, I'd prefer to do it here. If not, we should open an issue for this and make it a high priority after the 2023 data is merged.

Should be fixed now! I wonder if there's an easy way to test these in CI?

src/ferc_xbrl_extractor/datapackage.py
```diff
@@ -447,30 +458,53 @@ class Datapackage(BaseModel):
     def from_taxonomies(
         cls, taxonomies: dict[str, Taxonomy], db_uri: str, form_number: int = 1
     ) -> "Datapackage":
-        """Construct a Datapackage from an XBRL Taxonomy.
+        """Construct a Datapackage from parsed XBRL taxonomies.
+
+        FERC regularly releases new versions of their XBRL taxonomies, meaning
+        data from different years coforms to slightly different structures. This
```
e-belfer, line 464:
```suggestion
        data from different years conforms to slightly different structures. This
```

README.rst
```diff
@@ -124,7 +124,7 @@ filings, use the command:
 
 .. code-block:: console
 
-    $ xbrl_extract examples/ferc1-2021-sample.zip ./ferc1-2021-sample.sqlite \
+    $ xbrl_extract examples/ferc1-2021-sample.zip --db-path ./ferc1-2021-sample.sqlite \
```
e-belfer, line 127: @zschira are you able to run this successfully? I get:

```
 File "./ferc-xbrl-extractor/src/ferc_xbrl_extractor/instance.py", line 425, in instances_from_zip
 filing["filename"]: datetime.datetime.fromisoformat(
    ~~~~~~^^^^^^^^^^^^
TypeError: string indices must be integers, not 'str'
```

e-belfer:
> Hmm how were you checking the validation? We have an integration test, and validate the datapackage before writing to a json file.

I just ran `frictionless validate datapackage.json` locally. I suspect this is a similar issue to the one in #153, which noted that our integration test wasn't catching this problem?

I get the following error for each table:
> Please provide "dialect.sql.table" for reading

e-belfer (approved):
I have a ~2~ outstanding questions about the `frictionless` issue ~and the README examples / our sample data~, but it's not clear to me if either are a new problem or not. Those outstanding, everything looks good to go to me. I'm going to ask that if you don't deal with them in this PR, we open new issues for both of them.

README.rst
zschira: Did you do a `git pull`? I also had to update the example filings because they were the old format. When I run with the new filings everything works

e-belfer: Oh silly, that did it. Sorry about that!

PR #368, opened 2025-09-16
Output to duckdb and sqlite

zschira:
# Overview

This PR updates XBRL extractor to output to duckdb as well as SQLite. The purpose of this change is to make the raw FERC XBRL data available in Eel Hole.

I had initially started working on an update to only output duckdb, but this requires updating all places in PUDL where we access the SQLite files. This isn't too difficult, as we can switch the extraction to pull from the new duckdb files, but considering this touches the nightly builds, CI, and local development, it will be easier to just output both for now, and we can eventually fully deprecate the SQLite output.

# Testing

I executed the example command in the `README`, but added an option to also output duckdb.

`xbrl_extract examples/ferc1-2021-sample.zip -t examples/ferc1-xbrl-taxonomies.zip -f 6 --duckdb-path ferc1.duckdb --sqlite-path ferc1.sqlite`

### DIFF
--- a/src/ferc_xbrl_extractor/cli.py
+++ b/src/ferc_xbrl_extractor/cli.py
@@ -7,6 +7,7 @@ import sys
 from pathlib import Path
 
 import coloredlogs
+import duckdb
 from sqlalchemy import create_engine
 
 from ferc_xbrl_extractor import helpers, xbrl
@@ -23,10 +24,16 @@ def parse():
         type=Path,
     )
     parser.add_argument(
-        "-d",
-        "--db-path",
+        "--sqlite-path",
         default="ferc-xbrl.sqlite",
-        help="Store data in sqlite database specified in argument",
+        help="Path to SQLite DB to write extracted XBRL data.",
+        type=str,
+    )
+    parser.add_argument(
+        "--duckdb-path",
+        default=None,
+        help="Path to duckdb DB to write extracted XBRL data.",
+        type=str,
     )
     parser.add_argument(
         "-s",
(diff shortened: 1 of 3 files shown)


src/ferc_xbrl_extractor/cli.py
```diff
@@ -99,11 +106,38 @@ def parse():
...
+        if not data.empty:
+            # Write to SQLite
+            with engine.begin() as sqlite_conn:
+                data.to_sql(table_name, sqlite_conn, if_exists="append")
+
+            # Write to duckdb if passed a path to duckdb
+            if duckdb_path is not None:
+                with duckdb.connect(duckdb_path) as duckdb_conn:
+                    duckdb_conn.execute(
+                        f"CREATE TABLE {table_name} AS SELECT * FROM data"  # noqa: S608
```
zaneselvans, line 131: Is there a reason that the DuckDB version doesn't need the same check for an existing table, and the option to append if it exists that the SQLite output has?

jdangerx: It doesn't seem like we would ever expect to write to a table twice in one run, and also I don't think I've ever run this without `--clobber` (and then not immediately re-run with `--clobber`)... Maybe we should remove the clobber flag and just run `data.to_sql(..., if_exists="replace")` for the SQLite and `CREATE TABLE OR REPLACE` for the DuckDB?

src/ferc_xbrl_extractor/cli.py
```diff
@@ -99,11 +106,38 @@ def parse():
...
 
+def load_extracted(
+    extracted: xbrl.ExtractOutput,
+    sqlite_uri: str,
+    duckdb_path: str | None,
+    clobber: bool,
+) -> None:
+    """Write extracted data to SQLite/Duckdb databases."""
+    engine = create_engine(sqlite_uri)
+    if clobber:
```
jdangerx, line 117: If we decide to keep `clobber` around we should probably clobber the duckdb path too - or make that `CREATE TABLE` conditionally become `CREATE TABLE OR REPLACE`.

src/ferc_xbrl_extractor/cli.py
```diff
@@ -99,11 +106,38 @@ def parse():
...
+
+    for table_name, data in extracted.table_data.items():
+        # Loop through tables and write to database
+        if not data.empty:
+            # Write to SQLite
+            with engine.begin() as sqlite_conn:
+                data.to_sql(table_name, sqlite_conn, if_exists="append")
+
+            # Write to duckdb if passed a path to duckdb
+            if duckdb_path is not None:
```
jdangerx, line 128: nit: slightly annoying to have such deep nesting + conditionals here - it might be cleaner to have `write_to_sql` and `write_to_duckdb` as separate functions that get called in `main`. But really doesn't matter.

jdangerx (changes requested):
Let's talk about the clobbering-or-not behavior - otherwise good to go!

src/ferc_xbrl_extractor/cli.py
zschira: Decided to just remove the `clobber` option. No need for extra complexity we're not making use of.

PR #388, opened 2025-11-19
Fix duckdb outputs

zschira:
# Overview

This PR adds improved dtype handling, and fixes a bug causing tables in the new duckdb outputs to be missing columns.

# What problem does this address?
In [this](https://github.com/catalyst-cooperative/pudl/pull/4770) pr in PUDL, @zaneselvans  discovered that the new duckdb outputs were encountering a data type error, and the outputs seemed significantly smaller than the corresponding SQLite outputs. This PR fixes both of these problems.

# Testing

I've updated an integration test to compare extracted tables in the SQLite/ducdkb outputs to verify that they are identical. I've also run the new extractor against actual FERC1 data to recreate the error identified in the above PR, and demonstrate that it no longer occurs. This can be done with the following command (assuming you have the latest FERC1 archive downloaded in your PUDL workspace):

```
xbrl_extract ${PUDL_INPUT}/ferc1/10.5281-zenodo.17563678/ferc1-xbrl-2022.zip -t ${PUDL_INPUT}/ferc1/10.5281-zenodo.17563678/ferc1-xbrl-taxonomies.zip -f 1 --duckdb-path ferc1.duckdb -w 4
```

Running this command fails on `main`, and passes on this branch.

### DIFF
--- a/tests/integration/datapackage_test.py
+++ b/tests/integration/datapackage_test.py
@@ -96,6 +96,7 @@ def _create_schema(instant=True, axes=None):
                     'cid_1,EID1,filing,2023-01-01T00:00:01,2021-01-01,2021-12-31,"value 1","value 2",\n'
                     'cid_4,EID1,filing,2023-01-01T00:00:01,2020-01-01,2020-12-31,"value 3","value 4",\n'
                 ),
+                dtype="string",
                 parse_dates=["publication_time"],
             ),
         ),
@@ -109,6 +110,7 @@ def _create_schema(instant=True, axes=None):
                     'cid_4,EID1,filing,2023-01-01T00:00:01,2020-01-01,2020-12-31,total,"value 3","value 4",\n'
                     'cid_5,EID1,filing,2023-01-01T00:00:01,2020-01-01,2020-12-31,"Dim 1 Value","value 9","value 10",\n'
                 ),
+                dtype="string",
                 parse_dates=["publication_time"],
             ),
         ),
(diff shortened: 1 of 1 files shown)


src/ferc_xbrl_extractor/datapackage.py
```diff
@@ -391,9 +391,7 @@ def __init__(self, schema: Schema, period_type: str):
         """Create FactTable and prepare for constructing dataframe."""
         self.schema = schema
         # Map column names to function to convert parsed values
-        self.columns = {
-            field.name: CONVERT_DTYPES[field.type_] for field in schema.fields
-        }
+        self.columns = {field.name: field.type_ for field in schema.fields}
```
zschira: Change this to map column names to dtype labels like `"string"`, or `"integer"`. This allows us to use this dictionary to get both types to parse values, and pandas dtypes.

tests/integration/console_scripts_test.py
```diff
@@ -95,15 +78,13 @@ def test_extract_example_filings(script_runner, tmp_path, test_dir):
     assert ret.success
 
     # Sanity check the sqlite/duckdb outputs
-    sqlite_uri = f"sqlite:///{sqlite_path.absolute()}"
-    sqlite_engine = create_engine(sqlite_uri)
     with (
-        sqlite_engine.begin() as sqlite_conn,
+        duckdb.connect(sqlite_path) as sqlite_conn,
```
zschira: `duckdb` is happy to read SQLite, so we can just use `duckdb` for everything. This is what allowed me to delete the `_get_sqlite_tables` method.

zaneselvans (approved):
Glad this was relatively straightforward. Checking that the schemas of the two database outputs are the same seems like a good validation check. I think I will bring it over into the PUDL repo after releasing a new version of this package so we can have some confidence that the two versions of output are the same across all of the actual XBRL.

- I ran the sample command on main and it failed from year dtype issue.
- I ran the sample command from the PR and it worked.

PR #392, opened 2025-11-25
Fix datetime type for publication date

zschira:
# Overview

In [this pr](https://github.com/catalyst-cooperative/pudl/pull/4782) it was discovered that certain column types had changed due to #388. Most of these changes seem to be fixes where we are now using the correct dtype, however `datetime` columns across the board changed from `Timestamp` to `String`, which is incorrect. This PR fixes how we handle `datetime`s so we should now have appropriate dtypes.

### DIFF
--- a/src/ferc_xbrl_extractor/helpers.py
+++ b/src/ferc_xbrl_extractor/helpers.py
@@ -2,6 +2,20 @@
 
 import logging
 
+import pandas as pd
+
+
+def parse_dates(date_str: str) -> pd.Timestamp:
+    """Helper to normalize date strings/parse in a consistent way."""
+    try:
+        if "24:00:00" in date_str:
+            return pd.to_datetime(
+                date_str.replace("24:00:00", "00:00:00")
+            ) + pd.Timedelta(days=1)
+        return pd.to_datetime(date_str)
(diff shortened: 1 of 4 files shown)


src/ferc_xbrl_extractor/helpers.py
```diff
@@ -2,6 +2,20 @@
...
+def parse_dates(date_str: str) -> pd.Timestamp:
+    """Helper to normalize date strings/parse in a consistent way."""
+    try:
+        if "24:00:00" in date_str:
+            return pd.to_datetime(
+                date_str.replace("24:00:00", "00:00:00")
+            ) + pd.Timedelta(days=1)
+        return pd.to_datetime(date_str)
+    except ValueError:
+        return pd.NaT
```
zaneselvans: This is nice and compact.

Do you understand how we were getting valid timestamps before without doing any of this munging?

zschira: I think that the only column that actually was being properly turned into a `datetime` is the `publication_time` because that gets parsed separately from the RSS feed metadata, not XBRL directly. Other date columns were just getting written as strings to SQLite.

PR #428, opened 2026-05-13
Don't build resource or table for null instant or duration tables

cmgosnell:
# Overview

Working on https://github.com/catalyst-cooperative/eel-hole/issues/66

- before we made every instant and duration table. that was fine but we need a way to only display the tables that have any values in them. instead of just removing the tables from the parquet `datapackage.json` I decided to just go upstream and never make them in the first place.

- when there are no columns, don't make a resource and don't add said resource to the db schema.

# Testing

How did you make sure this worked? How can a reviewer verify this?
- [x] copied an old version of a `datapackage.json` from old all the tables method w/ old all the parquet files
   - [x]  verify that new method is dropping tables
   - [x] verify that new method drops tables that are empty parquet files
- [x] installed editable version of this repo in pudl.
- [x] ran the xbrl assets in the `ferc_to_sqlite` job
- [x] ran the ferc1 transforms w/in pudl and the validation tests...
   - [x]  this did require a small change over in pudl land

# To-do list

- [ ] add other TODO items here if necessary! questions that need to answered, decisions that need to be made, tests that need to be run, etc.
- [ ] Update relevant documentation - like comments, docstrings, README, release notes, etc.
- [ ] Review the PR yourself and call out any questions or issues you have

### DIFF
--- a/src/ferc_xbrl_extractor/datapackage.py
+++ b/src/ferc_xbrl_extractor/datapackage.py
@@ -310,6 +310,7 @@ class Resource(BaseModel):
         cls, fact_table: LinkRole, period_type: str, db_uri: str
     ) -> "Resource" | None:
         """Generate a Resource from a fact table (defined by a LinkRole).
+
         If the fact table is empty, i.e. there are no data columns, return None.
 
         Args:
--- a/src/ferc_xbrl_extractor/datapackage.py
+++ b/src/ferc_xbrl_extractor/datapackage.py
@@ -311,6 +311,7 @@ class Resource(BaseModel):
     ) -> "Resource" | None:
         """Generate a Resource from a fact table (defined by a LinkRole).
         If the fact table is empty, i.e. there are no data columns, return None.
+
         Args:
             fact_table: Link role which defines a fact table.
             period_type: Period type of table.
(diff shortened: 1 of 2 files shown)


tests/integration/datapackage_test.py
```diff
@@ -53,9 +53,9 @@ def test_datapackage_generation(test_dir, data_dir):
 
     all_tables = datapackage.get_fact_tables()
 
-    # 366 was just the value we had - this assertion is more of a regression
+    # 255 was just the value we had - this assertion is more of a regression
     # test than a normative statement
-    assert len(all_tables) == 370
+    assert len(all_tables) == 255
```
jdangerx, line 58: Just checked s3 and there *are* 255 tables so I believe this is working hooray

```
> s5cmd ls 's3://pudl.catalyst.coop/nightly/ferc1_xbrl/*.parquet' | wc -l
255
```

src/ferc_xbrl_extractor/datapackage.py
jdangerx, line 313:
```suggestion
        If the fact table is empty, i.e. there are no data columns, return None.
```

jdangerx:
```suggestion
    ) -> "Resource" | None:
```

jdangerx (approved):
Sweet, glad this was such a short diff - good job past @zschira for architectural foresight :)

Couple nits about docs/type annotations, but all non-blocking.

Plus, a gentle reminder that you'll need to do the version update etc, but you know that :)

PR #444, opened 2026-07-20
FIx easy type checking issues and related bugs

zaneselvans:
# Overview

See the updated release notes and README contents for additional documentation of what changed.

- After updating some repo infrastructure in #442 I tried to add type checking, and discovered there were quite a few issues, so I held off on trying to add that.
-This PR fixes the straightforward type annotation issues, and also a couple of bugs that the type checking highlighted.
- Many lines are `ty:ignore` comments due to Arelle being untyped -- we can add local stubs for the tiny subset of the Arelle API that we actually use, but that's a different PR.
- One of those bugs had to do with how we were constructing XBRL instances -- previously you could (in theory) build them from a ZipFile, a directory of filings, or an individual filing. But quite a while ago we realized you actually need the rssfeed metadata to differentiate between duplicate / updated records and also a stashed Taxonomy to apply to a whole batch of records. So the "directory of filings" and "single filing" instantiation paths have been broken for ages because they didn't offer those inputs, and we weren't actually ever using them in production. I removed the single-filing case, and attempted to fix the "directory of filings" case by assuming that the directory is exactly what you'd get if you unzipped one of our archives (including `rssfeed`).
- Another latent bug revealed by type checking was a case in which we assumed there would be a DuckDB path even if one hadn't been supplied, leading to a "no such method on None" error. Now we assume a default if one isn't provided. 
- Removes some 🧟 code `get_pandas_type()` which hasn't been used since 2022.
- Adds some new unit tests for things that weren't covered previously.
- Swaps a `df.apply()` for constructing `df.from_records()` in the hopes of speeding things up a little.
- Fixes what I think was a bug where we never escaped from an exponential backoff loop until `max_retries` because we had a `continue` where there should have been a `break`

In digging around a little more it seems like there may be some code that we don't actually use anywhere -- including the bit that had the `break` vs. `continue` bug -- since we never read Taxonomies from a URL and always assume we have archived inputs on disk. It might be worth doing an audit to see what code in here we actually use in either the `pudl-archiver` or `pudl` repos, and then deciding if we want to keep any of the code that we never use for completeness, etc. or if it's just cruft that we should trim.

# Testing

- All the unit and integration tests pass.
- Linting and formatting and type checking also pass.

# To-do list

- [x] Update relevant documentation - like comments, docstrings, README, release notes, etc.
- [x] Review the PR yourself and call out any questions or issues you have

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -253,7 +253,7 @@ relative_files = true
 precision = 2
 sort = "miss"
 skip_empty = true
-fail_under = 96
+fail_under = 99
 # fail_under is enforced once, after unit and integration coverage are combined, by
 # the final "coverage report" invocation in the `all` hatch script under
 # [tool.hatch.envs.test.scripts] above -- not here, since pytest-cov would otherwise
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -253,7 +253,7 @@ relative_files = true
 precision = 2
 sort = "miss"
 skip_empty = true
-fail_under = 95
+fail_under = 96
 # fail_under is enforced once, after unit and integration coverage are combined, by
 # the final "coverage report" invocation in the `all` hatch script under
 # [tool.hatch.envs.test.scripts] above -- not here, since pytest-cov would otherwise
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -94,6 +94,14 @@ types = [
     "ty>=0.0.61", # Extremely fast Rust-based Python type checker, from Astral
 ]
 
+[tool.uv]
+# Always use a uv-managed Python distribution -- never one discovered on the system
+# (Homebrew, apt, pyenv, etc.) -- and fetch it automatically if the pinned version
+# (see .python-version) isn't installed yet. Keeps local dev and CI reproducible
+# without relying on actions/setup-python or whatever happens to be on PATH.
+python-preference = "only-managed"
+python-downloads = "automatic"
+
 [tool.hatch.version]
 source = "vcs"
 
@@ -109,6 +117,13 @@ packages = ["src/ferc_xbrl_extractor"]
 [tool.hatch.envs.default]
 description = "Default environment with all development dependencies"
 features = ["dev", "docs", "tests", "types"]
+# Delegate venv creation and package installation to uv rather than pip/virtualenv.
+# Hatch still resolves *which* interpreter to hand uv (see the `python` pin below),
+# but once resolved, this makes uv do the actual venv creation and installs, and lets
+# HATCH_PYTHON (set to a `uv python find` path in CI) bypass Hatch's own PATH
+# discovery entirely -- see AGENTS.md and .github/workflows/*.yml.
+installer = "uv"
+python = "3.14"
 
 [tool.hatch.envs.test]
 description = "Run tests with pytest and collect test coverage"
@@ -135,6 +150,10 @@ all = ["coverage-erase", "unit", "integration", "coverage-report"]
 description = "Run linters and formatters"
 features = ["tests"]
 detached = true
+# `detached` makes this its own template root, so it doesn't inherit `installer`/
+# `python` from [tool.hatch.envs.default] -- repeat them here.
+installer = "uv"
+python = "3.14"
 
 [tool.hatch.envs.lint.scripts]
 check = ["ruff check ./", "ruff format --check ./"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -142,7 +142,7 @@ unit = "pytest --cov=src/ferc_xbrl_extractor --cov=tests --cov-append --cov-fail
 # tests parallelized this way each cap their own internal extraction concurrency
 # (workers=1) rather than also spawning a per-test process pool -- see comments in
 # tests/integration/data_quality_test.py and console_scripts_test.py.
-integration = "pytest -n auto --cov=src/ferc_xbrl_extractor --cov=tests --cov-append --cov-fail-under=0 --no-cov-on-fail tests/integration {args}"
+integration = "pytest -n auto --durations 10 --cov=src/ferc_xbrl_extractor --cov=tests --cov-append --cov-fail-under=0 --no-cov-on-fail tests/integration {args}"
 coverage-report = ["coverage report", "coverage xml"]
 all = ["coverage-erase", "unit", "integration", "coverage-report"]
 
--- a/src/ferc_xbrl_extractor/arelle_interface.py
+++ b/src/ferc_xbrl_extractor/arelle_interface.py
@@ -13,7 +13,35 @@ from pydantic import BaseModel
 
 
 def _taxonomy_view(taxonomy_source: str | FileSource.FileSource, max_retries: int = 7):
-    """Actually use Arelle to get a taxonomy and its relationships."""
+    """Use Arelle to load a taxonomy and build its parent-child relationship view.
+
+    As it parses a taxonomy, Arelle downloads the schema/linkbase files it
+    references to a local on-disk cache (managed by Arelle's own ``webCache``),
+    then reads them back from there. When two or more callers load the *same*
+    taxonomy concurrently -- e.g. multiple threads or processes each extracting
+    a different filing that shares one taxonomy version -- their cache writes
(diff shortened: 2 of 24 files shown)


src/ferc_xbrl_extractor/cli.py
```diff
@@ -148,6 +151,10 @@ def run_main(
     coloredlogs.install(fmt=log_format, level=loglevel, logger=logger)
 
     sqlite_uri = f"sqlite:///{sqlite_path.absolute()}"
+    # Default to writing DuckDB output alongside the SQLite output, unless the user
+    # asked for some other duckdb path.
+    if duckdb_path is None:
+        duckdb_path = sqlite_path.with_suffix(".duckdb")
```
zaneselvans, line 157: New default behavior here.

src/ferc_xbrl_extractor/instance.py
```diff
@@ -41,10 +42,14 @@ def from_xml(cls, elem: Element) -> "Period":
         if instant is not None:
             return cls(instant=True, end_date=instant.text)
 
+        start_date_elem = elem.find(f"{{{XBRL_INSTANCE}}}startDate")
+        end_date_elem = elem.find(f"{{{XBRL_INSTANCE}}}endDate")
+        assert start_date_elem is not None
+        assert end_date_elem is not None
```
zaneselvans, line 48: Check that these are not None before passing into the constructor.

src/ferc_xbrl_extractor/instance.py
```diff
@@ -148,12 +156,14 @@ class Context(BaseModel):
...
+        assert period_elem is not None
         return cls(
-            **{
-                "c_id": elem.attrib["id"],
-                "entity": Entity.from_xml(elem.find(f"{{{XBRL_INSTANCE}}}entity")),
-                "period": Period.from_xml(elem.find(f"{{{XBRL_INSTANCE}}}period")),
-            }
+            c_id=elem.attrib["id"],
+            entity=Entity.from_xml(entity_elem),
+            period=Period.from_xml(period_elem),
```
zaneselvans, line 166: Assert not none and pass in the elements directly rather than unpacking the dictionary.

src/ferc_xbrl_extractor/instance.py
```diff
@@ -437,6 +455,21 @@ def instances_from_zip(instance_path: Path | io.BytesIO) -> list[InstanceBuilder
         for filers_metadata in filings_metadata.values()
         for filing in filers_metadata
     }
+    return publication_times, taxonomy_versions
+
+
+def instances_from_zip(instance_path: Path | io.BytesIO) -> list[InstanceBuilder]:
```
zaneselvans, line 461: Split the two paths (zip archive vs. directory of files) into two functions.

src/ferc_xbrl_extractor/taxonomy.py
```diff
@@ -37,23 +37,7 @@ class XBRLType(BaseModel):
...
-        """
-        if self.base == "string" or self.base == "date" or self.base == "duration":
-            return "string"
-        if self.base == "decimal":
-            return "Float64"
-        if self.base == "gyear" or self.base == "integer":
-            return "Int64"
-        if self.base == "boolean":
-            return "boolean"
-        return None
```
zaneselvans, line 56: This hasn't been used since 2022, so I removed it. This came up looking at the test coverage and then I dug through the commit history.

src/ferc_xbrl_extractor/xbrl.py
```diff
@@ -25,9 +26,16 @@
 ExtractOutput = namedtuple("ExtractOutput", ["table_defs", "table_data", "stats"])
 
 
+class BatchResult(TypedDict):
+    """Dataframes and metadata extracted from one batch of instances."""
+
+    dfs: dict[str, pd.DataFrame]
+    metadata: dict[str, dict]
```
zaneselvans, line 33: Impose slightly more structure on this collection of outputs.

src/ferc_xbrl_extractor/arelle_interface.py
```diff
@@ -14,20 +13,48 @@
...
     cntlr = Cntlr.Cntlr()
     cntlr.startLogging(logFileName="logToPrint")
     model_manager = ModelManager.initialize(cntlr)
     for try_count in range(max_retries):
         try:
-            cntlr.logger.debug(f"Try #{try_count}: {taxonomy_source=}")
+            cntlr.logger.debug(f"Try #{try_count}: {taxonomy_source=}")  # ty:ignore[unresolved-attribute] -- pre-existing gap
             taxonomy = ModelXbrl.load(model_manager, taxonomy_source)
-            continue
+            break
```
zaneselvans, line 52: @zschira Maybe I'm understanding this wrong, but it seems like we want to `break` rather than `continue` here.  If we **haven't** raised an exception at this point, then it means we've got the taxonomy, and can proceed, right? With `continue` wouldn't we keep iterating through the loop up until we get to `max_retries`, even if a taxonomy has been found? If so then this seems like it might be a source of slowness when we're loading taxonomies generally.

src/ferc_xbrl_extractor/arelle_interface.py
```diff
@@ -14,20 +13,48 @@
 
 
 def _taxonomy_view(taxonomy_source: str | FileSource.FileSource, max_retries: int = 7):
-    """Actually use Arelle to get a taxonomy and its relationships."""
+    """Use Arelle to load a taxonomy and build its parent-child relationship view.
```
zaneselvans, line 16: I was confused about why this function was necessary and what it was doing, so expanded the docstring after I figured it out.

pyproject.toml
```diff
@@ -127,14 +142,18 @@ unit = "pytest --cov=src/ferc_xbrl_extractor --cov=tests --cov-append --cov-fail
 # tests parallelized this way each cap their own internal extraction concurrency
 # (workers=1) rather than also spawning a per-test process pool -- see comments in
 # tests/integration/data_quality_test.py and console_scripts_test.py.
-integration = "pytest -n auto --cov=src/ferc_xbrl_extractor --cov=tests --cov-append --cov-fail-under=0 --no-cov-on-fail tests/integration {args}"
+integration = "pytest -n auto --durations 10 --cov=src/ferc_xbrl_extractor --cov=tests --cov-append --cov-fail-under=0 --no-cov-on-fail tests/integration {args}"
```
zaneselvans, line 145: Report the durations of the slowest tests so we can see what the bottlenecks are clearly.

pyproject.toml
```diff
@@ -253,7 +272,7 @@ relative_files = true
 precision = 2
 sort = "miss"
 skip_empty = true
-fail_under = 95
+fail_under = 99
```
zaneselvans, line 275: 💯

zaneselvans (commented):
Almost all of the changed lines of code in here are fixes to types / type annotations or `# ty:ignore` comments for the Arelle API which we would need to generate stubs for (easy to do, but a different PR...). 

I've tried to highlight the handful of changes that were bugs which came up through the process of type checking or trying to debug why the tests were so crazy slow.

src/ferc_xbrl_extractor/arelle_interface.py
zschira: Yeah this should definitely be `break`ing

src/ferc_xbrl_extractor/instance.py
```diff
@@ -89,7 +94,7 @@ def from_xml(cls, elem: Element) -> "Axis":
             )
 
         if elem.tag.endswith("typedMember"):
-            dim = elem.getchildren()[0]
+            dim = elem[0]
```
zschira, line 97: I assume this change is equivalent?

src/ferc_xbrl_extractor/instance.py
zschira: Splitting these is definitely cleaner

zschira (approved):
Real type checking is nice!

src/ferc_xbrl_extractor/instance.py
zaneselvans: Yeah, the elem.getchildren() method was deprecated:

https://lxml.de/apidoc/lxml.etree.html#lxml.etree.CommentBase.getchildren

